PR #4, opened 2024-09-23
few more algorithm updates

danpape:
few more algorithm updates


danpape:
merging myself--not a good practice but want to get this update in before the presentation.

PR #11, opened 2025-02-25
fixed a few terminology definiton and referencing issues, and few typos too

danpape:


chapters/Terminology.md
wip-abramson, line 1: My understanding of what this was doing is creating an alias for a term. Such that we can reference either ::beacon:: ::beacon's:: or ::beacons:: and have it refer to the same term. Not, adding new terms into the terminology section.

@KDean-Dolphin can you confirm this.

jenniemeier: @wip-abramson and @danpape this is correct! We have several areas in the spec that should link to the Beacon definition, but we've used a plural, possessive, or other version of the term, so the added options in the terminology section allow the linking to refer to the original definition. The aliases don't show up in the published terminology list, just on our end for editing.

danpape: Got it - I didn't realize that was happening! I'll back out those changes asap.

jenniemeier: We'll make sure to mention it in the did:btc1 call tomorrow. I only know because I worked with Kevin on adding links throughout the spec. Super cool pandoc feature, but definitely one we should highlight and explain :)

README.md
```diff
@@ -26,4 +29,4 @@ you will need to rerun the script `npm run pandoc-spec-local` to see the changes
 Included under the `notebooks` folder are a set of Jupyter notebooks that implement the
 various features of the **did:btc1** specification. These are included as helpful reference
 material for those intending to implement the specification. To run the notebooks locally see
-the `noteooks/README.md`.
\ No newline at end of file
+the `noteooks/README.md`.
```
wip-abramson, line 32:
```suggestion
the `notebooks/README.md`.
```

PR #39, opened 2025-03-18
Update Read algorithm based on implementation experience

wip-abramson:
This update is based on implementing the Read algorithm by connecting to a local btc polartest node.


wip-abramson:
I have fixed the test vectors based on changes to the bip340 di spec updates. I think this is ready to merge. Lets discuss briefly in our call.

PR #40, opened 2025-03-18
Update errors during Read operation

peacekeeper:
This updates the parse algorithm to check for the correct DID method name before parsing the rest of the identifier, and to raise a "methodNotSupported" error instead of "invalidDid" if the method is not "btc1".

Also updates the "InvalidDID" error to "invalidDid" in some places, and adds links to the DID Resolution spec.


wip-abramson (approved):

jintekc:
> cc @danpape and @jintekc.
> 
> I think we discussed this on a call. Whether we should handle other cases that just the `invalidDid` error. Markus makes a good point here. Not sure how that we change our use of RegEx, but we should add the `methodNotSupported` error case handling.

FWIW - `@web5/dids` has some handy code that incorporates many of these errors into a single did-error class. I've been using this in my implementation to handle various error cases: https://github.com/decentralized-identity/web5-js/blob/main/packages/dids/src/did-error.ts

Issue #14, opened 2025-02-26
Update did:btc1 identifier syntax


wip-abramson:
Currently the identifier syntax is as follows: 

did-btc1 = "did:btc1:" [ version ":" ] [ network ":" ] id-bech32

We want to change this to be

did-btc1 = "did:btc1:" id-bech32

Where the `id-bech32` value encapsulates the version and network data of the identifier.

We will use four HRP values: (x, k, tx, tk). 

- (x,k) for mainnet identifiers (external, key-based).
- (tx,tk) for testnet identifiers.

We need to figure out how the data part of the bech32 should be parsed.

It will be some bytes that contains: <version><network><genesisBytes>

Any thoughts about the best approach?

Are the values of version and network optional? Are there defaults? 

Is network needed/allowed for mainnet identifiers?

wip-abramson:
Thanks for putting this together @KDean-Dolphin. I like the proposal, lets discuss it in next weeks call with Ryan. 

The only thing I would add is to:

> The start of the identifier remains as is, i.e., "did:btc1:k1..." or "did:btc1:x1...".

We felt it would still be useful for a human to eyeball a btc1 identifier and tell if it is a production DID or not. To do this we discussed adding two additional hrp values: `tk` and `tx`. 

Still doesn't tell you which network it is on, but you can know if it is mainnet or not.

Also, Bitcoin doesn't call its primary network `mainnet` but `bitcoin`. So we should probably adopt that.

danpape:
This is a great proposal Dean. I'm sorry to comment so late, but I had a question:

Let's imagine that at some point in the future, let's say version 4 of the identifier syntax, we suddenly decide "hey, we really need another extra byte or two in between the identifier byte and the key data in order to support new feature XYZ." Do you feel that this proposal of yours would be able to extend to that case? Or would we need some other mechanism?

KDean-Dolphin:
@wip-abramson Regarding the human-readable...

Bech32 encodes five bits at a time, so we can interpret the version and network by looking at the first few characters. How that manifests depends on the HRP.

[...]

PR #48, opened 2025-03-26
Rewrote syntax per discussion of #14

KDean-Dolphin:


danpape:
I think this all looks great. Thank for the proposal and writing it all up.

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+
+| Decoded bytes     | Version | Network         |
+|-------------------|--------:|-----------------|
+| *00* 03 c7 ...    |       1 | bitcoin         |
+| *23* 03 c7 ...    |       3 | testnet v3      |
+| *C9* 03 c7 ...    |      13 | custom signet 2 |
+| *F2 90* 03 c7 ... |      18 | custom signet 2 |
+| *FF 72* 03 c7 ... |      38 | regtest         |
+
+### Encoding a **did:btc1** Identifier
```
wip-abramson, line 98: Hmm, I am still on the fence as to whether this should go here or not. 

Perhaps its fine, but this algorithm is only going to be called from the Create section of the CRUD operations.

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+  * "hash"
+* `version` - number
+* `network` - one of:
+  * "bitcoin"
+  * "signet"
+  * "regtest"
+  * "testnet3"
+  * "testnet4"
+  * number
+* `idBytes` - byte array, one of:
```
wip-abramson, line 113:
```suggestion
* `genesisBytes` - byte array, one of:
```

I think I would prefer to use this name as consistent with the rest of the spec.

chapters/Syntax.md
wip-abramson: If we are keeping it here, we still need to integrate this algorithm with the CRUD operations. I.e. the Create should call out to encode. 

I think this should be done in this PR

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+Decode the **did:btc1** identifier as follows:
+
+1. Split `id` into an array of `components` at the colon `:` character.
+1. If the length of the `components` array is not `3`, raise `InvalidDID` error.
+1. If `components[0]` is not "did", raise `InvalidDID` error.
+1. If `components[1]` is not "btc1", raise `InvalidDID` error.
+1. Set `encodedString` to `components[2]`.
+1. Pass `encodedString` to the Bech32m decoding algorithm, retrieving `hrp` and
+   `decodedBytes`.
+1. If the Bech32m decoding algorithm fails, raise `InvalidDID` error.
```
wip-abramson, line 166:
```suggestion
1. If the Bech32m decoding algorithm fails, raise `invalidDid` error.
```

Matches the DID spec - https://w3c.github.io/did-resolution/#invaliddid

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+1. Map `hrp` to `idType` from the following:
+   1. "k" - "key"
+   1. "x" - "hash"
+   1. other - raise `InvalidDID` error
```
wip-abramson, line 170:
```suggestion
   1. other - raise `invalidDid` error
```

wip-abramson: Change should be done throughout

chapters/Syntax.md
wip-abramson: Open to a discussion as to by idBytes is better. 

I just see the idBytes as the whole thing. Wheras the genesisBytes are the seed for the ID

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+Given:
+
+* `id` - a ***did:btc1** identifier
+
+Decode the **did:btc1** identifier as follows:
+
+1. Split `id` into an array of `components` at the colon `:` character.
+1. If the length of the `components` array is not `3`, raise `InvalidDID` error.
+1. If `components[0]` is not "did", raise `InvalidDID` error.
+1. If `components[1]` is not "btc1", raise `InvalidDID` error.
```
wip-abramson, line 162:
```suggestion
1. If `components[1]` is not "btc1", raise `methodNotSupported` error.
```

I think https://w3c.github.io/did-resolution/#methodnotsupported.

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+   to `nibbles.length / 2 - 1` and `encodingBytes[index] = (nibbles[2 * index]
+   << 8) | nibbles[2 * index + 1]`.
+1. Append `idBytes` to `encodingBytes`.
+1. Set `id` to "did:btc1:".
+1. Pass `hrp` and `decodedBytes` to the Bech32m encoding algorithm, retrieving
+   `encodedString`.
+1. Append `encodedString` to `id`.
+1. Return `id`.
+
+### Decoding a **did:btc1** Identifier
```
wip-abramson, line 151: Again, can you update the Read algorithms to call this algorithm and remove the duplicate / conflicting steps

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+   next nibble (the lower nibble of the current byte or the higher nibble of the
+   next byte) and return to the previous step.
+1. If `version` is not `1`, raise `InvalidDID` error.
+1. Advance to the next nibble and set `networkValue` to its value.
+1. Map `networkValue` to `network` from the following:
+   1. `0` - "bitcoin"
+   1. `1` - "signet"
+   1. `2` - "regtest"
+   1. `3` - "testnet3"
+   1. `4` = "testnet4"
```
wip-abramson, line 187:
```suggestion
   1. `4` - "testnet4"
```

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+    1. "testnet3" - `3`
+    1. "testnet4" - `4`
+1. If `network` is a number, append `network + 7` to `nibbles`.
+1. If the number of entries in `nibbles` is odd, append `0`.
+1. Create a `decodedBytes` byte array from `nibbles`, where `index` is from `0`
+   to `nibbles.length / 2 - 1` and `encodingBytes[index] = (nibbles[2 * index]
+   << 8) | nibbles[2 * index + 1]`.
+1. Append `idBytes` to `encodingBytes`.
+1. Set `id` to "did:btc1:".
+1. Pass `hrp` and `decodedBytes` to the Bech32m encoding algorithm, retrieving
```
wip-abramson, line 146: There is currently a section in the appendix for the bech32 encode and decode algorithms.

We should either reference them or remove them.

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+
+It is sometimes useful to differentiate between production (bitcoin network) and
+test (other network) **did:btc1** identifiers without having to go through the
+decoding process.
+
+Bech32 encodes five bits at a time, so the version and network (8 bits, assuming
+version is 15 or below) can be interpreted by looking at the first two
+characters. How that manifests depends on the HRP.
 
-### Examples
```
wip-abramson, line 29: I think we should still show some kind of examples. But we can add those later.

wip-abramson:
This looks great, thanks @KDean-Dolphin! 

I think the main comments I made above are to do with making sure this change aligns and ties in to the rest of the specification.

chapters/Syntax.md
KDean-Dolphin: I'll go with the majority opinion. If we agree that the Syntax chapter is OK, it won't take long to replace all the did:btc1 strings with updated values and to update references to the encoding/decoding.

I would prefer to keep the encoding/decoding in the Syntax chapter, just from a logical grouping perspective. Having "here's what it looks like, here's how to create it, and here's how to interpret it" in one place is more logical, in my opinion, even though the "create" part is referenced only once.

chapters/Syntax.md
KDean-Dolphin: I hate it, but given that it's the way the spec is written, we're stuck with it.

chapters/Syntax.md
KDean-Dolphin: The examples that were deleted showed the equivalence between different DIDs, which doesn't apply anymore. The document will have more comprehensive examples in subsequent chapters.

chapters/CRUD-Operations.md
```diff
@@ -84,57 +73,16 @@ in `resolutionOptions` or is set to the ::Resolution Time:: of the request.
 
 To do so it executes the following algorithm:
 
-1. Let `identifierComponents` be the result of running the algorithm in
-   [Parse **did:btc1** identifier], passing in the `identifier`.
+1. Pass `identifier` to the [did:btc1 Identifier Decoding](#didbtc1-identifier-decoding)
```
wip-abramson, line 76: Why not just return a struct containing the 4 properties?

wip-abramson (approved):
Great, thanks!

chapters/CRUD-Operations.md
KDean-Dolphin: I prefer the style of "this is a list of inputs" and "this is a list of outputs". Having "this is a list of inputs" and "this is a struct output with a list" is a legacy of the fact that functions can take multiple parameters and return only one result.

I intend to go through the document with a fine tooth comb after I get back from the Bitcoin Programming course next week. After working through it and creating examples, I have some ideas about reordering that I think would help, and refactoring the algorithms for consistency is part of that, but it's beyond the scope of this PR.

danpape: I agree with Kevin, I think the spec should be pretty generic about saying "these are all the inputs and these are all the outputs" and implementers in whatever language can hide them in objects or not as they desire according to their own language's idioms.

wip-abramson:
Okay, I think this is good to merge. @KDean-Dolphin if you can just resolve the conflicts and we can get it in.

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+  * "testnet4"
+  * number
+* `genesisBytes` - required, byte array, one of:
+  * a compressed secp256k1 public key if `idType` is "key"
+  * a hash of an initiating external DID document if `idType` is "external"
+
+Encode the **did:btc1** identifier as follows:
+
+1. If `idType` is not a valid value per above, raise `invalidDid` error.
+1. If `version` is not `1`, raise `invalidDid` error.
```
wip-abramson, line 120: This test is strange to me. Esp given the check below.

`1. Set `fCount` equal to `(version - 1) / 15`, rounded down.`

Are we saying that this is the alogrithm to encode a v1 identifier only. If we are then do we need the above step?

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+    1. "signet" - `1`
+    1. "regtest" - `2`
+    1. "testnet3" - `3`
+    1. "testnet4" - `4`
+1. If `network` is a number, append `network + 7` to `nibbles`.
+1. If the number of entries in `nibbles` is odd, append `0`.
+1. Create a `dataBytes` byte array from `nibbles`, where `index` is from `0` to
+   `nibbles.length / 2 - 1` and `encodingBytes[index] =
+   (nibbles[2 * index] << 8) | nibbles[2 * index + 1]`.
+1. Append `genesisBytes` to `encodingBytes`.
```
wip-abramson, line 144:
```suggestion
1. If the number of entries in `nibbles` is odd, append `0`.
1. Create a `dataBytes` byte array from `nibbles`, where `index` is from `0` to
   `nibbles.length / 2 - 1` and `dataBytes[index] =
   (nibbles[2 * index] << 8) | nibbles[2 * index + 1]`.
1. Append `genesisBytes` to `dataBytes`.
```

encodingBytes or dataBytes. I think I prefer dataBytes

wip-abramson, line 144: I dont think append is the right language here. At least as implementing it I needed something different.

My genesisBytes are a set of bytes like `b'\x02\xb8j-\x9a6K\xad\xaf\xb9\xc8A\xbd\xd2\xa6\x8b\x9c\xc66CE*\xbf\x18\x8c2\xe1\x15\r\x9e\x9c\x89B'`

I think we need to convert those into a bytearray.

The concatenate the two byte arrays. In python this is just `b1 + b2`

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+    1. "bitcoin" - `0`
+    1. "signet" - `1`
+    1. "regtest" - `2`
+    1. "testnet3" - `3`
+    1. "testnet4" - `4`
+1. If `network` is a number, append `network + 7` to `nibbles`.
+1. If the number of entries in `nibbles` is odd, append `0`.
+1. Create a `dataBytes` byte array from `nibbles`, where `index` is from `0` to
+   `nibbles.length / 2 - 1` and `encodingBytes[index] =
+   (nibbles[2 * index] << 8) | nibbles[2 * index + 1]`.
```
wip-abramson, line 143:
```suggestion
   (nibbles[2 * index] << 4) | nibbles[2 * index + 1]`.
```

I believe

wip-abramson (changes requested):
I was working on an implementation against these new spec changes and spotted a few bugs that we should fix.

I have a working implementation of the encode. Will try get one for decode too.

chapters/Syntax.md
```diff
@@ -1,36 +1,225 @@
...
+1. Set `version` to `1`.
+1. If at any point in the remaining steps there are not enough nibbles to
+   complete the process, raise `invalidDid` error.
+1. Start with the first nibble (the higher nibble of the first byte) of
+   `dataBytes`.
+1. Add the value of the current nibble to `version`.
+1. If the value of the nibble is hexadecimal `F` (decimal `15`), advance to the
+   next nibble (the lower nibble of the current byte or the higher nibble of the
+   next byte) and return to the previous step.
+1. If `version` is not `1`, raise `invalidDid` error.
```
wip-abramson, line 180: Again, this seems odd to me. We do all this work to figure out the version. Then raise an error if its not 1.

We could just raise an error if the high nibble is not 0x0

wip-abramson:
My implementation for encode and decode is here - https://github.com/LegReq/did-btc1-py/blob/main/libbtc1/did.py

chapters/Syntax.md
KDean-Dolphin: The code is intended to work with any version, with minimal changes to the specification. So, for version 17, assuming no changes to the layout, the only line that would have to change is this one. However, I'll rewrite it to make it a little easier to change in future.

> If `version` is greater than `1`, raise `invalidDid` error.

chapters/Syntax.md
KDean-Dolphin: I prefer "append", because it's explicit about where it goes. It shouldn't happen, but "concatenate x and y" can be interpreted as x+y or y+x. The order is implied, and I can't imagine it being misinterpreted, but I've been proven wrong before...

chapters/Syntax.md
KDean-Dolphin: Same reply; future-proofing.

wip-abramson:
LGTM, merging. 

@danpape , @peacekeeper, @jintekc - just a note we will need to refactor existing implementations to match. 

I have started on some test cases here - https://github.com/LegReq/did-btc1-py/blob/main/test/test_did.py

Still need to add some failure states.

PR #62, opened 2025-04-16
Address minor bugs in deterministic doc generation algs

wip-abramson:
This:

- Adds the controller property to the DID doc
- Removes the uncessary data integrity spec from context (included in DID core I think)
- Uses full DID URLs, instead of relative URLs

### DIFF
--- a/chapters/CRUD-Operations.md
+++ b/chapters/CRUD-Operations.md
@@ -111,14 +111,14 @@ returns a `initialDocument`.
 1. Set `keyBytes` to `identifierComponents.genesisBytes`.
 1. Initialize a `initialDocument` variable as an empty object.
 1. Set `initialDocument.id` to the `identifier`.
+1. Set `initialDocument.controller` to the `identifier`
 1. Initialize a `contextArray` to empty array:
     1. Append the DID Core context "https://www.w3.org/ns/did/v1".
-    1. Append the Data Integrity context "https://w3id.org/security/data-integrity/v2".
     1. Append a **did:btc1** context.
     1. Set `initialDocument['@context]' to contextArray`.
 1. Create an initial verification method:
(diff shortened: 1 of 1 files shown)


peacekeeper (approved):
Yes, agree with all these changes.

peacekeeper:
> Removes the uncessary data integrity spec from context (included in DID core I think)

Actually, no. The term "Multikey" is not included in `https://www.w3.org/ns/did/v1`. I think we have three options:

1. Add `https://w3id.org/security/multikey/v1` as an additional context to define "Multikey".
2. Use `https://www.w3.org/ns/did/v1.1`, which will have "Multikey", see: https://github.com/w3c/did/blob/main/contexts/v1.1rc1
3. Use a custom `did:btc1` specific context. We may have to do that anyway, in order to define terms such as `SingletonBeacon`.

Sorry, maybe I approved too fast.

wip-abramson:
My sense is that we should be using the 1.1 context. But you are right, we are using a custom btc1 context already for SingletonBeacon etc.

chapters/CRUD-Operations.md
```diff
@@ -111,14 +111,14 @@ returns a `initialDocument`.
 1. Set `keyBytes` to `identifierComponents.genesisBytes`.
 1. Initialize a `initialDocument` variable as an empty object.
 1. Set `initialDocument.id` to the `identifier`.
+1. Set `initialDocument.controller` to the `identifier`
 1. Initialize a `contextArray` to empty array:
     1. Append the DID Core context "https://www.w3.org/ns/did/v1".
```
wip-abramson, line 116:
```suggestion
    1. Append the DID Core v1.1 context "https://www.w3.org/ns/did/v1.1".
```

Issue #75, opened 2025-06-15
OP_PUSH32 or OP_PUSHBYTES_32 ?


peacekeeper:
In [Find Next Signals](https://dcdpr.github.io/did-btc1/#find-next-signals), the spec says:

> Filter the beaconSpends, identifying all transactions whose 0th transaction output is of the format [OP_RETURN, OP_PUSH32, <32bytes>].

But I find that in our implementation, we use a regex pattern like this:

```
^OP_RETURN OP_PUSHBYTES_32 ([0-9a-fA-F]{64})$
```

Can someone explain why it's `OP_PUSHBYTES_32` instead of `OP_PUSH32` ? Maybe a mistake in the spec ?

Issue #69, opened 2025-05-28
OP_RETURN Vout Position


jintekc:
During implementation, @wip-abramson and I came across a nuance regarding the placement of a DID Update Payload OP_RETURN, and we wanted to open it up for discussion.

## Context

The spec defines that the update payload hash should be the `0th transaction output` (see [Find Next Signals](https://github.com/dcdpr/did-btc1/blob/main/chapters/CRUD-Operations.md#find-next-signals)). Defining the Vout index in the spec makes the resolvers job easier avoiding the need to test each Vout to find the `OP_RETURN` reducing complexity further (see #56).

> 1. Set `signals` to an empty array.
> 1. For each `beacon` in `beacons`:
>    1. Set `beaconSpends` to the set of all Bitcoin transactions on the specified 
>    `network` that spend at least one transaction input controlled by the `beacon.address`
>    with a blockheight greater than or equal to the `contemporaryBlockheight`.
>    1. Filter the `beaconSpends`, identifying all transactions whose 0th transaction output 
>    is of the format `[OP_RETURN, OP_PUSH32, <32bytes>]`.
>    1. For each of the filtered `beaconSpends` push the following `beaconSignal` object onto the `signals` array.
> 1. If `signals` is empty, return `signals`.
> 1. Sort `signals` by `blockheight` from lowest to highest.
> 1. Set `nextSignals` to all `signals` with the lowest `blockheight`.
> 1. Return `nextSignals`.

We discovered this nuance when Will failed to verify a recent DID Update Payload transaction I submitted [5fadd5b82d60f36cc750610860c2420a5b1266d81b1096bb981a01f4c1c09809](https://mutinynet.com/tx/5fadd5b82d60f36cc750610860c2420a5b1266d81b1096bb981a01f4c1c09809 ) where I put the OP_RETURN in the last Vout position.

Initially, I was putting the `OP_RETURN` in the first (0th-index) Vout postitiong, but after looking at various blocks on mutinynet, I realized the following:
1. The coinbase of every newly mined block contains an `OP_RETURN` as the last Vout
2. Additional instances of OP_RETURN Vout in non-coinbase txs on mutinynet are also the last Vout
3. Making the OP_RETURN the 0th Vout points the lookup of a `prevTx` to 1 and it seemed like 0 made more sense (intuitively)

[...]

PR #79, opened 2025-06-20
Rename OP code for accuracy and update signal txout index to last

wip-abramson:
This PR addresses issues #75 and #69.


danpape:
Looks good

Issue #73, opened 2025-06-10
Update all errors to align with DID resolution adoption of Problem Details


wip-abramson:
Currently in the spec we are using the old error codes specified by DID core. E.g. `invalidDid`.

We want to update these to align with DID resolutions use of Problem Details, see https://w3c.github.io/did-resolution/#errors

wip-abramson:
@peacekeeper wondering if you would be able to take a look at this issue. Be good to be aligned with the DID Resolution spec, but think we are a ways from this in the current spec text.

wip-abramson:
This is a duplicate of #103

PR #84, opened 2025-07-10
Added section on file sharing with details on IPFS mapping

KDean-Dolphin:


chapters/File-Sharing.md
```diff
@@ -0,0 +1,92 @@
...
+
+All files in this specification, except for deterministically generated initial
+DID documents, are identified by their SHA256 cryptographic hash calculated
+according to the [JSON Canonicalization and Hash] algorithm. As a resolver goes
+through the resolution process, it encounters one or more document hashes, which
+it uses to identify the files of interest. There are two ways in which these
+files may be shared.
+
+While it's possible for a single **did:btc1** identifier to mix the two file
+sharing mechanisms, it is not recommended.
```
wip-abramson, line 11: I think we should say why here. Or maybe you do later?

chapters/File-Sharing.md
```diff
@@ -0,0 +1,92 @@
...
+
+In short, when a resolver is presented with a **did:btc1** identifier, it is
+also presented with files matching the SHA256 hashes it encounters during the
+resolution process. If any SHA256 hash doesn't have a corresponding file, the
+resolution fails.
+
+### Content-Addressable Storage (CAS)
+
+Content-Addressable Storage (CAS) is a mechanism by which a file is stored
+based on its content, not its name or location. The content address is
```
wip-abramson, line 29:
```suggestion
Content-Addressable Storage (CAS) is a mechanism by which a file stored is addressed by its content, not its name or location. The content address is
```

chapters/File-Sharing.md
```diff
@@ -0,0 +1,92 @@
...
+determined by a cryptographic hash of the file. The hash is then passed into a
+retrieval function specific to the type of CAS to retrieve the file.
+
+Any CAS that provides a deterministic mapping from a SHA256 hash of a file may
+be used, and a resolver SHOULD be informed of the specific CAS mechanism so that
```
wip-abramson, line 34: So this is fine. And I think I agree, but we need to define an approach for how this can be done. I am not sure the spec has this yet.

chapters/File-Sharing.md
```diff
@@ -0,0 +1,92 @@
...
+it can retrieve documents associated with a **did:btc1** identifier efficiently.
+If the CAS mechanism is not provided, the resolver MAY iterate through all
+supported CAS mechanisms to find the files or it MAY return with an error
```
wip-abramson, line 37: This also is fine, but lets define the error similar to here - https://w3c.github.io/did-resolution/#errors

Probably we need our own errors section, I think did:btc1 is going to define a few

chapters/File-Sharing.md
```diff
@@ -0,0 +1,92 @@
...
+Leaves option.
+
+The IPFS CIDv1 is a binary identifier constructed from the file hash as:
+
+* 0x01, the code for CIDv1;
+* 0x00, the [multicodec](https://github.com/multiformats/multicodec) code for 
+raw binary;
+* 0x12, the [multihash](https://github.com/multiformats/multihash) code for
+SHA-256; and
+* the SHA-256 of the file.
```
wip-abramson, line 85: an example could be useful here

chapters/File-Sharing.md
jintekc: An approach for "how a resolver retrieves a file from CAS" sounds like an implementation detail, but maybe I'm not fully understanding.

It seems like section [4.2.1.2.2 CAS Retrieval
](https://dcdpr.github.io/did-btc1/#cas-retrieval) does a very specific job of outlining how to retrieve from IPFS specifically. If anything, we may need to abstract that section slightly to be more generally inclusive of CAS systems and the maybe a sub section for "IPFS Retrieval."

jintekc (approved):

chapters/File-Sharing.md
KDean-Dolphin: I can change it to:

...it is not recommended, as it may complicate file management and error analysis.

Basically, having two ways of managing the files for a DID is a recipe for bugs. It can be done, and the algorithms as I've laid them out don't care, but it means more debugging when a verifier can't find a file.

chapters/File-Sharing.md
KDean-Dolphin: I don't think there's a way to define a generic approach. IPFS is the only CAS we know of that has a SHA256 mapping, so we can be specific about that, but any other CAS would have to have its own documentation.

chapters/File-Sharing.md
KDean-Dolphin: I need to review the errors in both this and the Beacons rewrite. I'll take care of it.

pandoc-spec.options.json
```diff
@@ -8,6 +8,7 @@
     "Introduction-and-Motivation.md",
     "Terminology.md",
     "Syntax.md",
+    "File-Sharing.md",
```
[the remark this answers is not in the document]
KDean-Dolphin: We may be able to move some to an appendix, but the concept of sidecar especially is new and deserves discussion in the main part of the spec. It's not like the Merkle tree appendix, which aggregates concepts from multiple sources and walks through a detailed "why" of privacy. I think the position is logical.

Syntax -> File Sharing -> Beacons

chapters/File-Sharing.md
wip-abramson: I think also the benefit of using IPFS is resolution is possible from just the did:btc1 identifier. As soon as you mix approaches to have sidecar, then this is never true again

chapters/File-Sharing.md
wip-abramson: I think we should create a different issue to explore Nostr. I think it could be interesting, but don't know enough about it. Dont think it currently supports content addressing, but surely it could

chapters/File-Sharing.md
wip-abramson: Great thanks Kevin. There is an issue #73, perhaps you could do all this in one PR?

pandoc-spec.options.json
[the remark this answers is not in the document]
wip-abramson: Okay, fair. Lets leave it in, then we can do a full review later and see if it feels right. I guess my mental model is just the other specs generally start with the CRUD. 

But I do like where you are coming from. First explain the peices that CRUD relies on. THen explain the CRUD

chapters/File-Sharing.md
wip-abramson: Resolving this as we can handle it in a separate PR

wip-abramson:
@KDean-Dolphin could you also address this: `TODO: Is this right? Are implementations just supposed to check all CAS they trust?` At https://dcdpr.github.io/did-btc1/#fetch-content-from-addressable-storage

Maybe it just needs cleaning up?

chapters/File-Sharing.md
KDean-Dolphin: True except for SMTBeacon, which always requires an externally-provided proof of (non-)inclusion.

wip-abramson:
Okay merging this. 

I see this has a section called Sidecar. I wonder if this is where we define the Sidecar Data. My sense is no, we want a stand alone section that defines Sidecar Data. This would include defining `sidecarPackage` that would be sent to the DID resolver in the resolution options. As well as how a resolver can process that package to turn it into a sidecarDocumentMap and a set of sidecarProofs

See #178

PR #99, opened 2025-07-30
top level controller property no longer used by the spec

wip-abramson:


jintekc:
Fully agree and align with this change. Approved.

Issue #36, opened 2025-03-11
Review Terminology section


danpape:
There were several issues in the old tracker that dealt with questioning whether certain definitions and terms in our Terminology section were clear enough or or needed extra references. I just want to note that in this issue so that we don't forget to do that at some point. I will add some of the original issues description as comments below.

danpape:
For example:

We make reference to Z-Caps and some of the Z-Cap terminology without defining them ourselves. Is it OK to just point to the Z-Cap spec's Terminology section? (https://w3c-ccg.github.io/zcap-spec/#terminology)

wip-abramson:
@danpape can I assign this one to you?

I am thinking that we should also add Decentralized Identifier (DID) and DID document to the terminology section. Maybe some more

wip-abramson:
I think adding Intermediate DID document would be useful

wip-abramson:
This is related to the Beacons Refactor and an alignment of terms #86

PR #118, opened 2025-08-06
Beacons refactor

KDean-Dolphin:
This is an overhaul of the Update-Beacons.md (now Beacons.md) file. There's work to be done to integrate it into the rest of the document, and there's a list of TODOs at the top of Beacons.md that I still need to address, but there's enough here for a discussion of the changes and eventual merge.

The introductory documentation has been significantly updated, including a discussion of the attributes for advertising a cohort set, and forward references have been removed.

- SingletonBeacon
    - No changes to the concept, but documentation significantly improved.
- CIDAggregateBeacon
    - Renamed to MapBeacon (aggregate is implied).
    - Removed dependency on IPFS (map file can be delivered by sidecar).
- SMTAggregateBeacon
    - Renamed to SMTBeacon (aggregate is implied).
    - Documentation expanded.

For discussion:

- Map Beacon uses a key-value pair mechanism, where the key is the DID and the value is the hash of the DID update payload. This exposes all the DIDs participating in the cohort. An alternative is an indexed approach, where the key is the hash of the DID, which hides the other participating DIDs. Options are:
    - Keep as is (DID as key).
    - Add indexing as an option, with a top-level property defining the key type (DID or index).
    - Eliminate the "DID as key" option and go with an obfuscated index only.
- There's no library available to do what we need for the SMT Beacon. I can create one in Python and TypeScript and demonstrate interoperability.
- The original version of the documentation included signal metadata in the return, but I'm hard pressed to see the value. The only thing the aggregator needs to share with the cohort participants is the signal ID.
- We should add documentation about transaction settling time, i.e., how long we should wait before we consider the transaction truly immutable.
- We should add documentation about recovering from aggregator cohort dissolution, i.e., what to do when a cohort participant leaves and the cohort is no longer viable.

Let's discuss on the next sync call.


wip-abramson:
@KDean-Dolphin re the nonce generation phase. @jintekc just shared this excellent doc - https://github.com/chainx-org/chainx-technical-archive/blob/main/Daiwei/Taproot/04_Basic%20principles%20of%20Musig2.md

We may want to cite this somewhere.

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+payloads is made by broadcasting it on a Bitcoin network. A ::Beacon:: is
+identified by a Bitcoin address and emits ::Beacon Signals:: by broadcasting
+valid Bitcoin transactions that spend from this address.
+
+A ::Beacon:: is included as a service in DID documents, with the Service
+Endpoint identifying a Bitcoin address to watch for ::Beacon Signals::. All
+::Beacon Signals:: broadcast from this ::Beacon:: MUST be processed as part of
+resolution. The type of the ::Beacon:: service in the DID document defines how
+::Beacon Signals:: MUST be processed.
+
```
jintekc, line 25: Add an example for clarity.
> ```json
> { 
>   "service": [
>         {
>             "type": "SingletonBeacon",
>             "id": "did:btc1:k1qypcylxwhf8sykn2dztm6z8lxm43kwkyzf07qmp9jafv3zfntmpwtks9hmnrw#initialP2PKH",
>             "serviceEndpoint": "bitcoin:moZUuvjVnipp1jd6y6a2vdhjLjdaWpuTCv"
>         },
>         {
>             "type": "SingletonBeacon",
>             "id": "did:btc1:k1qypcylxwhf8sykn2dztm6z8lxm43kwkyzf07qmp9jafv3zfntmpwtks9hmnrw#initialP2WPKH",
>             "serviceEndpoint": "bitcoin:tb1qtqaktvg6nc9yajx07ukwcus47pz9hvsk84vqsh"
>         },
>         {
>             "type": "SingletonBeacon",
>             "id": "did:btc1:k1qypcylxwhf8sykn2dztm6z8lxm43kwkyzf07qmp9jafv3zfntmpwtks9hmnrw#initialP2TR",
>             "serviceEndpoint": "bitcoin:tb1p5ss9d8e4rtehk32ldjtdpm38vj29yx3gwuad94q6zpx3udk8nh0q58zeeq"
>         }
>     ]
> }
> ```

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+* the hash of a [Merkle tree](https://en.wikipedia.org/wiki/Merkle_tree) root,
+  where each leaf node is deterministically selected by a **did:btc1**
+  identifier and contains a hash associated with the **did:btc1** identifier.
+
+When defining a service for a ::Beacon:::
+
+* `type` is one of "SingletonBeacon", "MapBeacon", or "SMTBeacon"
+* `serviceEndpoint` is a Bitcoin address represented as a URI following the
+  [BIP21 scheme](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)
+
```
jintekc, line 63: Move the `When defining a service for a ::Beacon:::` paragraph up to be below the above DID Document example.

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+How the Bitcoin address and the cryptographic material that controls it are
+generated is left to the implementation.
+
```
jintekc, line 66: Move this section

> When defining a service for a ::Beacon:::
>
> * `type` is one of "SingletonBeacon", "MapBeacon", or "SMTBeacon"
> * `serviceEndpoint` is a Bitcoin address represented as a URI following the  [BIP21 scheme](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)
>
> How the Bitcoin address and the cryptographic material that controls it are generated is left to the implementation.

Up to be below the DID Doc example

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+Three types of ::Beacons:: are defined, with two of them aggregating multiple
+DID update payloads into a single transaction. How coordination between an
+aggregator and multiple DID controllers is managed is out of scope, but one
+possible mechanism is outlined in “MuSig2 3-of-3 Multisig with Coordinator
+Facilitation” at
+[MuSig2 Sequence Diagrams](https://developer.blockchaincommons.com/musig/sequence/#musig2-3-of-3-multisig-with-coordinator-facilitation).
+
+When defining a cohort set, the aggregator may define the conditions for the
+cohort, including but not limited to:
+
```
jintekc, line 90: What is a "cohort set"? This is the initial introduction of the combinations of these terms without little context or definition, i.e. "cohort" + "set". While there is a definition above for "cohort" that includes the term "set" I think it would be better to keep the term "cohort" consistent or better would be to make it very explicit in its naming.

When defining a Beacon Aggregation Cohort, the Beacon Aggregator may define the conditions for the cohort

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+::DID Update Payload:: targeting a single DID document. It creates a ::Beacon
+Signal:: that commits to a single DID update payload. This is done directly by
+the DID controller, and there is no cohort or aggregator.
+
+If the DID update payload is not publicly discoverable (i.e., is not published
+to a ::CAS:: under its hash), the only parties that are aware of it are the DID
+controller and any parties provided it by the DID controller.
+
+The `type` of the `service` for a Singleton Beacon is "SingletonBeacon".
+
```
jintekc, line 111: Insert example

```json
{
    "type": "SingletonBeacon",
    "id": "did:btc1:k1qypcylxwhf8sykn2dztm6z8lxm43kwkyzf07qmp9jafv3zfntmpwtks9hmnrw#initialP2PKH",
    "serviceEndpoint": "bitcoin:moZUuvjVnipp1jd6y6a2vdhjLjdaWpuTCv"
}
```

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+        loop For each DID in documents map...
+            alt Payload provided
+                A ->> A: Publish DID update payload to CAS
+            end
+        end
+    end
+```
+
+##### Construct Map Document
+
```
jintekc, line 266: "Map Document" feels like its comes out of nowhere. In fact, this is the first time that term is used in this way up to this point in the spec. Maybe "Construct Map Beacon Document" or just "Construct Map Beacon" would be better for mental logical flow.

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+    * "testnet3"
+    * "testnet4"
+    * "mutinynet"
+    * number
+* `serviceEndpoint` - required, a Bitcoin address represented as a URI
+* One and only one of, required:
+    * `didUpdatePayload` - the DID update payload whose SHA256 hash is to be
+      broadcast in the ::Beacon Signal::
+    * `didUpdateHashBytes` - the SHA256 hash in binary form of the DID update
+      payload to be broadcast in the ::Beacon Signal::
```
jintekc, line 129: What is the goal of this option between `didUpdatePayload` and `didUpdateHashBytes`? Is it to allow for different data types? If so, I'd recommend making it one arg `didUpdatePayload`  and saying "this can be a hex or binary sha256 hash".

To me, this introduces unnecessary confusion because up until this point, I've been thinking solely about `didUpdatePayload` and now this newly introduced term leads to questions about what that is and where has it been prior referenced. When reviewed, this is the first time the term is introduced.

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+an optimized Sparse Merkle Tree as defined in [Appendix - Optimized Sparse
+Merkle Tree Implementation] and publishes the Merkle root.
+
+An SMT Beacon provides maximum privacy for the DID controller, as the DID
+controller never has to reveal their DIDs or DID update payloads to the
+aggregator. This introduces a small risk, as the DID controller is not required
+to prove control over a DID in order to participate.
+
+The `type` of the `service` for an SMT Beacon is "SMTBeacon".
+
```
jintekc, line 377: Insert JSON example

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+controller and any parties provided it by the DID controller. However, any party
+that has access to or is provided the consolidated key-value pair document is at
+least aware of the existence of all **did:btc1** identifiers in the document and
+the existence of their DID update payloads.
+
+For a Map Beacon, proof of non-inclusion of a **did:btc1** identifier is simply
+its absence from the consolidated key-value pair document.
+
+The `type` of the `service` for a Map Beacon is "MapBeacon".
+
```
jintekc, line 181: Insert JSON example

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+1. If `documentsMap` contains a duplicate `did`, raise InvalidParameter error.
+1. Set `mapDocument` to the JSON representation of `documentsMap` in the form
+   `{"<did1>": "<hashString1>", "<did2>": "<hashString2>",...}` where:
+   1. `didN` is the *N*th `did` in `documentsMap`;
+   1. `hashBytesN` is the *N*th `didUpdateHashBytes` in `documentsMap` if
+      defined or the result of passing the *N*th `didUpdatePayload` in
+      `documentsMap` to the [JSON Canonicalization and Hash] algorithm; and
+   1. `hashStringN` is the hexadecimal string representation of `hashBytesN`.
+
```
jintekc, line 288: I think we need to discuss aligning terms in our upcoming meeting, so we can be sure to capture the meaning of these things and align the naming for congruence throughout the spec. Map Beacon, document map, consolidated map, did update hash bytes, etc. New terms are okay, so long as they all align sensibly and have proper definitions.

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+1. Ensure `bitcoinAddress` is funded; if not, fund this address.
+1. Initialize `unsignedSpendTx` to a Bitcoin transaction that spends a
+   transaction controlled by the `bitcoinAddress` and contains at least one
+   transaction output. This signal output MUST have the format
+   `[OP_RETURN, OP_PUSHBYTES32, <hashBytes>]`. If the transaction contains
+   multiple transaction outputs, the signal output MUST be the last transaction
+   output.
+
+##### Construct Peers Array
+
```
jintekc, line 512: I was initially completely thrown off by this header until I reviewed the UML diagram 2 or 3 times to realize that this is describing one of the steps in that diagram. I think there needs to be a better way to display this. Or a note referencing the diagram.

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+Actors in signaling are as follows:
+
+* DID controller - a party that controls one or more **did:btc1** identifiers
+  participating in a ::Beacon::
+* cohort - a set of DID controllers participating in a ::Beacon:: and in the
+  construction of the n-of-n P2TR Bitcoin address for the ::Beacon::
+* aggregator - the party responsible for aggregating updates from all DID
+  controllers and creating the ::Beacon Signals::
+* verifier - a party verifying a **did:btc1** identifier presentation
+
```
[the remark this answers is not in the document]
wip-abramson: I suggest updating the terms here. Then we can add them to the terminology section when handling #36

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+### Aggregation
+
+Three types of ::Beacons:: are defined, with two of them aggregating multiple
+DID update payloads into a single transaction. How coordination between an
+aggregator and multiple DID controllers is managed is out of scope, but one
```
wip-abramson, line 83:
```suggestion
Three types of ::Beacons:: are defined: SingletonBeacon, MapBeacon and SMTBeacon. 
Two of them, MapBeacon and SMTBeacon, support aggregation, i.e. the act of 
committing to multiple DID update payloads into a single transaction.
```

@jintekc I would encourage you to submit change requests like this. I like your suggestion, I tweaked it slightly (combining to committing)

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+Facilitation” at
+[MuSig2 Sequence Diagrams](https://developer.blockchaincommons.com/musig/sequence/#musig2-3-of-3-multisig-with-coordinator-facilitation).
+
+When defining a cohort set, the aggregator may define the conditions for the
+cohort, including but not limited to:
+
+* Automatic publication to ::CAS:: (Map Beacon only)
+* Minimum and/or maximum number of DID controllers
+* Minimum and/or maximum number of DIDs per DID controller
+* Cost of enrolment
```
wip-abramson, line 94:
```suggestion
* Cost of enrollment
```

chapters/Beacons.md
wip-abramson: Thanks @jintekc the examples are a great catch. @KDean-Dolphin can you add them as json files like the other examples

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+1. Return `signalId`.
+
+### Map Beacon
+
+A Map Beacon creates a ::Beacon Signal:: that commits to multiple DID update
+payloads, each identified by a **did:btc1** identifier. To do so, it constructs
+a new document of key-value pairs, where the key is the **did:btc1** identifier
+and the value is the hash of the corresponding DID update payload, and publishes
+a hash of the consolidated document.
+
```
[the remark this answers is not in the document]
wip-abramson: I like this suggestion apart from some terminology. coordinator -> Beacon Aggregator. Aggregated Update is DID Update Bundle I believe.

chapters/Beacons.md
wip-abramson: Hmm tricky. I don't like Construct Map Beacon. Because this is not the beacon. I was going to say Construct Map Beacon Signal. But its not that either.

This is the payload that the Beacon Signal will commit to.

wip-abramson: Maybe Construct Beacon Map?

Although really throughout the document we have been calling this thing a DID Update Bundle. Maybe thats a bad name and we should change it. But we should consistently name the thing.

Maybe it is a DID Update Commitment Map? A bit long winded

chapters/Beacons.md
wip-abramson: I agree. Although I am concerned that our meeting has a vastly oversubscribed agenda. 

@KDean-Dolphin could you join us on Tuesday at 7am Pacific?

Today I will try and put together a google doc outlining all the different things we need to name that this spec defines. This can hopefully drive the terminology update #36

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+the set of ::Beacons:: they use to broadcast updates to their DID document over
+time. Resolution of a DID MUST process signals from all ::Beacons:: identified in the
+latest DID document and apply them in the order determined by the version specified
+by the `didUpdatePayload`.
+
+All resolvers of **did:btc1** DIDs MUST support the Beacon Types defined in this
+specification.
+
+The commitment is a SHA256 hash that represents one of the following:
+
```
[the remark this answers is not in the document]
jintekc: My understanding of a commitment in this context (cryptography and blockchains) is some data (typically hashed) broadcasted via transaction and included in a time-bound block. I may be misunderstanding the nuancens there, however, since I'm more of a pragmatic study of the realm. That said, if my understanding is wrong, then it would make sense that my above rewording would also be wrong and not make sense. In which case, please forgive my ignorance :smile:

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+1. Broadcast `spendTx` on the Bitcoin network defined by `network`.
+1. Set `signalId` to the Bitcoin transaction identifier of `spendTx`.
+1. If `cas` and `didUpdatePayload` are defined, publish `didUpdatePayload` to
+   the ::CAS:: network defined by `cas`.
+1. Return `signalId`.
+
+### Map Beacon
+
+A Map Beacon creates a ::Beacon Signal:: that commits to multiple DID update
+payloads, each identified by a **did:btc1** identifier. To do so, it constructs
```
jintekc, line 165:
```suggestion
payloads, each identified by a **did:btc1** identifier. To do so, the Map Beacon Aggregator fills an empty Map object with identifiers of participants who have signaled their desired inclusion in a DID Update Bundle. Each key in the Map is a **did:btc1** identifier of a Beacon Participant pointing to a value of the hash of a DID update payload. The coordinator consolidates the Map into a single hash and publishes it to a CAS.

**Example Map Beacon Object**
```json
{ 
   "did:btc1:participantA": "participanthashA",
   "did:btc1:participantB": "participanthashB",
   "did:btc1:participantC": "participanthashC",
   "did:btc1:participantD": "participanthashD",
   "did:btc1:participantE": "participanthashE",
}
```
**Example Consolidated Map Beacon Hash**
`participanthashABCDE`
```

File: chapters/Beacons.md [hunk 18]
```diff
@@ -0,0 +1,725 @@
...
+        end
+    end
+```
+
+##### Construct Map Document
+
+Given:
+
+* `documentsMap` - required, a map of key-value pairs consisting of:
+    * `did` - required, a unique **did:btc1** identifier (key)
```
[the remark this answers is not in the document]
jintekc: Ok, so I think this should become `didUpdatePayloadMap`.

chapters/Beacons.md
jintekc: That's fair. I only suggested the live discussion in our meeting because I think @KDean-Dolphin said that he wanted to discuss this PR then.

chapters/Beacons.md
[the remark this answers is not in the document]
jintekc: Resolving in favor of the above suggestion edits.

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+* Minimum and/or maximum number of DID controllers
+* Minimum and/or maximum number of DIDs per DID controller
+* Cost of enrolment
+* Cost per signal per DID or DID controller
+* Minimum and/or maximum time between signals
+* Number of pending updates that trigger a signal
+
+### Singleton Beacon
+
+A ::Singleton Beacon:: is a ::Beacon:: that can be used to commit to a single
```
jintekc, line 101:
```suggestion
A ::Singleton Beacon:: is a ::Beacon:: used to create a ::Beacon Signal:: that commits to a single ::DID Update Payload::
```

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+::DID Update Payload:: targeting a single DID document. It creates a ::Beacon
+Signal:: that commits to a single DID update payload. This is done directly by
+the DID controller, and there is no cohort or aggregator.
+
```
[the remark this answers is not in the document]
jintekc: Resolving in favor of suggested edit aboveResolving in favor of suggested edit above

chapters/Beacons.md
jintekc, line 111:
```suggestion
```{.json include="json/Beacons/Singleton-service.json"}
```
```

Comment (jintekc) [reply chapters/Beacons.md:111]: I pushed some JSON example files, adding the above suggested edit and resolving this comment.

File: chapters/Beacons.md (hunk 7 above)
Comment (jintekc) [reply chapters/Beacons.md:129]: Agreed 💯 I think it we should keep didUpdatePayload, a SHA256 hash in hex string format.

File: chapters/Beacons.md [hunk 21]
```diff
@@ -0,0 +1,725 @@
...
+    * "regtest"
+    * "testnet3"
+    * "testnet4"
+    * "mutinynet"
+    * number
+* `serviceEndpoint` - required, a Bitcoin address represented as a URI
+* One and only one of, required:
+    * `didUpdatePayload` - the DID update payload whose SHA256 hash is to be
+      broadcast in the ::Beacon Signal::
+    * `didUpdateHashBytes` - the SHA256 hash in binary form of the DID update
```
jintekc, line 128:
```suggestion
 
```

chapters/Beacons.md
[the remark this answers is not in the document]
KDean-Dolphin: @jintekc That's exactly what we're doing. We're referencing the hash of a DID update payload (SingletonBeacon), the hash of a map of hashes (MapBeacon), or the root hash of an optimized sparse Merkle tree. Unless I'm missing something in your suggestion...

chapters/Beacons.md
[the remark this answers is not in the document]
KDean-Dolphin: @jintekc What about "map document" instead of "consolidated document"?

chapters/Beacons.md
KDean-Dolphin: The reason for supporting "one or the other" is to give the cohort participant the option of hiding their DID update payload from the aggregator (by providing the hash only). If I'm doing everything via sidecar, giving my DID update payloads to the aggregator introduces another point at which data about my DID can leak.

chapters/Beacons.md
KDean-Dolphin: @wip-abramson Sorry, I missed this. Add me to the next one, please.

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+How the Bitcoin address and the cryptographic material that controls it are
+generated is left to the implementation.
+
+### Actors
+
+Actors in signaling are as follows:
+
+* DID controller - a party that controls one or more **did:btc1** identifiers
+  participating in a ::Beacon::
+* cohort - a set of DID controllers participating in a ::Beacon:: and in the
```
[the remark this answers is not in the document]
wip-abramson: I like Beacon Cohort. @KDean-Dolphin thoughts? I am working on a separate terminology doc and will include this.

chapters/Beacons.md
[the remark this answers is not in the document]
wip-abramson: Currently in the spec we have:

> DID Update Bundle
> A JSON object that maps did:btc1 identifiers to [Content Identifiers](https://dcdpr.github.io/did-btc1/#def-content-identifier) (CIDs) that identify [DID Update Payloads](https://dcdpr.github.io/did-btc1/#def-did-update-payload) for the identified DID. DID Update Bundles are announced by CIDAggregate Beacons.

Happy to change this if we want. This is largely legacy terms from DID ion I believe. But Map Document feels a little generic.

Could be Update Map Document. 

Definitely agree with @jintekc here we need one term that we consistently are using throughout the spec for this data

chapters/Beacons.md
wip-abramson: @jintekc I am temped to say these examples are a separate PR. What do you think?

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+1. If `documentsMap` contains a duplicate `did`, raise InvalidParameter error.
+1. Set `mapDocument` to the JSON representation of `documentsMap` in the form
+   `{"<did1>": "<hashString1>", "<did2>": "<hashString2>",...}` where:
+   1. `didN` is the *N*th `did` in `documentsMap`;
+   1. `hashBytesN` is the *N*th `didUpdateHashBytes` in `documentsMap` if
+      defined or the result of passing the *N*th `didUpdatePayload` in
+      `documentsMap` to the [JSON Canonicalization and Hash] algorithm; and
+   1. `hashStringN` is the hexadecimal string representation of `hashBytesN`.
+
+##### Construct Unsigned Bitcoin Transaction
```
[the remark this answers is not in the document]
wip-abramson: @KDean-Dolphin

wip-abramson:
Still a fair bit of work to get this PR merged. But it is moving in the right direction.

@jintekc I think we defer all examples to a separate issue personally.

There is good discussion about terminology going on here, we should address that. But noting that we do not need to fix the terms consistently throughout, because that can happen in a separate pass throughout the whole doc.

@KDean-Dolphin I think Bryan is right that it would be useful to add some accompanying text to the mermaid diagrams

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+  identifier and contains a hash associated with the **did:btc1** identifier.
+
+When defining a service for a ::Beacon:::
+
+* `type` is one of "SingletonBeacon", "MapBeacon", or "SMTBeacon"
+* `serviceEndpoint` is a Bitcoin address represented as a URI following the
+  [BIP21 scheme](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)
+
+How the Bitcoin address and the cryptographic material that controls it are
+generated is left to the implementation.
```
jintekc, line 65:
```suggestion

```

chapters/Beacons.md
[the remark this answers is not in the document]
jintekc: Resolving and adding suggestion above

chapters/Beacons.md
rxgrant: > Is there a reason why you would not want to always hide your DID update payload from the aggregator

Yes, originally the CID_Aggregator was for public and institutional identity where the DID Controller wants the Aggregator to do the work of keeping it online for validators.  Controller is still (collectively) _responsible_ for pinning to IPFS (and individually responsible for sending sidecar as a last resort) if CID_Aggregator fails, but if CID_Aggregator is reputable then Controller doesn't have to be in the business of maintaining the accounts and bandwidth to do so.

The CID_Aggregator didn't have obsfucation options because Controller's use of a CID_Aggregator was the option - choosing a bundle of such tradeoffs.  You would never want to send ugly long maps sidecar (including the ones you weren't in) when you could just send an SMT proof.

If we continue with privacy options for MapBeacon, we need a clear motivation section.

chapters/Beacons.md
jintekc: Sure I tried to include them, but I can pull them out of the PR if that's we want to do

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+* the hash of a document of key-value pairs, where each key is a **did:btc1**
+  identifier and the value is the hash of a DID update payload or null for that
+  key; or
+* the hash of a [Merkle tree](https://en.wikipedia.org/wiki/Merkle_tree) root,
+  where each leaf node is deterministically selected by a **did:btc1**
+  identifier and contains a hash associated with the **did:btc1** identifier.
+
+When defining a service for a ::Beacon:::
+
+* `type` is one of "SingletonBeacon", "MapBeacon", or "SMTBeacon"
```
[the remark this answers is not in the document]
jintekc: Sorry, I may not have moved the content I'm suggesting for deletion up. I was trying to suggest we move all of this content

> When defining a service for a ::Beacon:::
> * `type` is one of "SingletonBeacon", "MapBeacon", or "SMTBeacon"
> * `serviceEndpoint` is a Bitcoin address represented as a URI following the [BIP21 scheme](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)
>
> How the Bitcoin address and the cryptographic material that controls it are generated is left to the implementation.

up to be just underneath

> A ::Beacon:: is included as a service in DID documents, with the Service Endpoint identifying a Bitcoin address to watch for ::Beacon Signals::. All ::Beacon Signals:: broadcast from this ::Beacon:: MUST be processed as part of resolution. The type of the ::Beacon:: service in the DID document defines how ::Beacon Signals:: MUST be processed.
> When defining a service for a ::Beacon:::
> * `type` is one of "SingletonBeacon", "MapBeacon", or "SMTBeacon"
> * `serviceEndpoint` is a Bitcoin address represented as a URI following the [BIP21 scheme](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)
>
> How the Bitcoin address and the cryptographic material that controls it are generated is left to the implementation.

KDean-Dolphin:
@wip-abramson It will be done by tomorrow.

KDean-Dolphin:
> @KDean-Dolphin re the nonce generation phase. @jintekc just shared this excellent doc - https://github.com/chainx-org/chainx-technical-archive/blob/main/Daiwei/Taproot/04_Basic%20principles%20of%20Musig2.md

I don't see anything there about nonce generation specifically. As long as the source is properly randomized, there's nothing special to nonce generation.

chapters/Beacons.md
KDean-Dolphin: I'm not sold on the use of this term, especially as there's no equivalent for the SMTBeacon. I've removed "document"; let's see how it looks once I've pushed the changes.

chapters/Beacons.md
KDean-Dolphin: There's logic in the section where the aggregator uploads to CAS if a) CAS is defined and b) the payload is provided by the Beacon Participant. Basically, the Beacon Participant is handing off that part of the work to the aggregator, as the aggregator already has to upload the map to CAS as well.

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+
+An SMT Beacon provides maximum privacy for the DID controller, as the DID
+controller never has to reveal their DIDs or DID update payloads to the
+aggregator. This introduces a small risk, as the DID controller is not required
+to prove control over a DID in order to participate.
+
+The `type` of the `service` for an SMT Beacon is "SMTBeacon".
+
+#### Create Cohort
+
```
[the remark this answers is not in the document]
KDean-Dolphin: Done.

chapters/Beacons.md
```diff
@@ -0,0 +1,727 @@
...
+   1. Set `id` to the hexadecimal string representation of `genesisBytes`.
+   1. Get `didDocument` from `sidecarDocumentsMap` by its `id` if available, or
+      from ::CAS:: by its `id` if not and `cas` is defined.
+   1. Update placeholder values in `didDocument` with `did` as required.
+1. Until terminated:
+   1. Set `didUpdatePayload` to null.
+   1. For each `service` in `didDocument.service` where `service.type` is
+      "SingletonBeacon", "Map Beacon", or "SMT Beacon":
+      1. Get the next transaction from the Bitcoin address at
+         `service.serviceEnpoint`.
```
danpape, line 633:
```suggestion
         `service.serviceEndpoint`.
```

chapters/Beacons.md
```diff
@@ -0,0 +1,727 @@
...
+      `didUpdatePayload`.
+   1. If `targetVersionId` is defined and `didDocument.versionId` =
+      `targetVersionId`, terminate.
+1. If `targetVersionId` is defined and `didDocument.versionId` ≠
+   `targetVersionId`, raise InvalidDidUpdate error.
+1. If `targetVersionId` is not defined:
+   1. For each `service` in `didDocument.service` where `service.type` is
+      "SingletonBeacon", "Map Beacon", or "SMT Beacon":
+      1. Get the next transaction from the Bitcoin address at
+         `service.serviceEnpoint`.
```
danpape, line 664:
```suggestion
         `service.serviceEndpoint`.
```

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+least aware of the existence of all **did:btc1** identifiers in the document and
+the existence of their DID update payloads.
+
+For a Map Beacon, proof of non-inclusion of a **did:btc1** identifier is simply
+its absence from the consolidated key-value pair document.
+
+The `type` of the `service` for a Map Beacon is "MapBeacon".
+
+#### Create Cohort
+
```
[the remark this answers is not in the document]
KDean-Dolphin: Done.

chapters/Beacons.md
KDean-Dolphin: Steps in the sequence diagram that refer to subsections are marked with "(*)", e.g., "Construct peers array (*)".

chapters/Beacons.md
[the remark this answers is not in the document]
KDean-Dolphin: Aligned with new terminology.

chapters/Beacons.md
KDean-Dolphin: Replaced per Will's suggestion.

chapters/Beacons.md
KDean-Dolphin: Accepted, with ending backticks.

chapters/Beacons.md
```diff
@@ -0,0 +1,725 @@
...
+* the hash of a single DID update payload associated with a **did:btc1**
+  identifier;
+* the hash of a document of key-value pairs, where each key is a **did:btc1**
+  identifier and the value is the hash of a DID update payload or null for that
+  key; or
+* the hash of a [Merkle tree](https://en.wikipedia.org/wiki/Merkle_tree) root,
+  where each leaf node is deterministically selected by a **did:btc1**
+  identifier and contains a hash associated with the **did:btc1** identifier.
+
+When defining a service for a ::Beacon:::
```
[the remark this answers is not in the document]
KDean-Dolphin: I'd like to keep the text, as it helps to break it from the bulleted list immediately above it.

chapters/Beacons.md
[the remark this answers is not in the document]
KDean-Dolphin: Moved per suggestion.

chapters/Beacons.md
```diff
@@ -0,0 +1,727 @@
...
+
+> **TODO**:
+> 
+> * Ensure that all `sidecarDocuments` and `smtProofs` have been used if
+>   `targetVersionId` is not defined.
+> * Review exception names.
+> * Replace "Validate that..." with explicit test and exception.
+> * Review need for anything more than signal ID to be shared after signaling.
+> * Review format of `service` entry, determine if `cas` should be supported.
+> * Review change from CIDAggregateBeacon to MapBeacon.
```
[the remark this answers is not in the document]
KDean-Dolphin: Removed; issue to be created shortly.

chapters/Beacons.md
KDean-Dolphin: Fixed.

chapters/Beacons.md
```diff
@@ -0,0 +1,587 @@
...
+* DID controller - A party that controls one or more **did:btc1** identifiers participating in a ::BTC1 Beacon::.
+* Beacon Cohort - The set of unique cryptographic keys participating in a ::BTC1 Beacon:: that make up its n-of-n MuSig2 Bitcoin address.
+* Beacon Aggregator - The entity that coordinates the protocols of an aggregate ::BTC1 Beacon::, specifically the "Create Beacon Cohort" and "Announce Beacon Signal" protocols.
+* Beacon Participant - A member of a Beacon Cohort, typically a DID controller, that controls cryptographic keys required to partially authorize the broadcasting of a Beacon Signal to the Bitcoin blockchain.
+* Verifier - A party verifying a **did:btc1** identifier presentation.
+
+### Aggregation
+
+Three types of ::BTC1 Beacons:: are defined: SingletonBeacon, MapBeacon and SMTBeacon.  Two of them, MapBeacon and SMTBeacon, support aggregation, i.e. the act of  committing to multiple ::BTC1 Beacon Announcements:: in a single transaction.
+
```
wip-abramson, line 48: I would like to find a way to support referencing the terms Map Beacon etc here. Apprecaite that the type string is MapBeacon. Open to suggestions for how best to handle this

chapters/Beacons.md
wip-abramson: No dont pull any out. Just lets have a pass that just focuses on examples

chapters/Beacons.md
```diff
@@ -0,0 +1,587 @@
...
+1. if `network` is a number and is outside the range of 1-4, raise InvalidParameter error.
+1. Set `bitcoinAddress` to the decoding of `serviceEndpoint` following BIP21.
+1. Ensure `bitcoinAddress` is funded; if not, fund this address.
+1. Initialize `unsignedSpendTx` to a Bitcoin transaction that spends a transaction controlled by the `bitcoinAddress` and contains at least one transaction output. This signal output MUST have the format `[OP_RETURN, OP_PUSHBYTES32, <hashBytes>]`. If the transaction contains multiple transaction outputs, the signal output MUST be the last transaction output.
+
+##### Construct Peers Array
+
+Given:
+
+* `smt` - required, optimized SMT
```
wip-abramson, line 435:
```suggestion
* `smt` - required, optimized ::SMT::
```

chapters/Beacons.md
```diff
@@ -0,0 +1,587 @@
...
+* `index` - required, index provided by DID controller
+
+Calculate the path to the root for the index:
+
+1. Set `peers` to empty array.
+1. Set `node` to leaf node for `index`.
+1. While `node` is not root of `smt`:
+   1. Set `parentNode` to parent of `node`.
+   1. If `node` is left of `parentNode`, add `{right: <rightHashString>}`to `peers`, where `rightHashString` is the hexadecimal string representation of the value at `parentNode.rightNode`.
+   1. If `node` is right of `parentNode`, add `{left: <leftHashString>}` to `peers`, where `leftHashString` is the hexadecimal string representation of the value at `parentNode.leftNode`.
```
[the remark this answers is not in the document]
wip-abramson: @jintekc

chapters/Beacons.md
```diff
@@ -0,0 +1,587 @@
...
+   1. Set `didUpdatePayload` to null.
+   1. For each `service` in `didDocument.service` where `service.type` is "BTC1Beacon":
+      1. Get the next transaction from the Bitcoin address at `service.serviceEndpoint`.
+      1. If `targetVersionTime` is defined and is less than the transaction time, skip to next `service`.
+      1. If the last output transaction is not of the form `[OP_RETURN, OP_PUSHBYTES32, <hashBytes>]`, skip to next `service`.
+      1. Extract `hashBytes` from the last output transaction.
+      1. If `service.beaconType` is not "SingletonBeacon", "MapBeacon", or "SMTBeacon", raise InvalidParameter error.
+      1. If `service.beaconType` is "SingletonBeacon", set `tempDidUpdatePayload` to the result of [Process Singleton Beacon Signal].
+      1. If `service.beaconType` is "MapBeacon", set `tempDidUpdatePayload` to the result of [Process Map Beacon Signal].
+      1. If `service.beaconType` is "SMTBeacon", set `tempDidUpdatePayload` to the result of [Process SMT Beacon Signal].
```
wip-abramson: This is fine for now. But I will note this is still under discussion and may need fixing later. Shouldnt stop this getting in

chapters/Beacons.md
```diff
@@ -0,0 +1,587 @@
...
+
+1. Set `peers` to empty array.
+1. Set `node` to leaf node for `index`.
+1. While `node` is not root of `smt`:
+   1. Set `parentNode` to parent of `node`.
+   1. If `node` is left of `parentNode`, add `{right: <rightHashString>}`to `peers`, where `rightHashString` is the hexadecimal string representation of the value at `parentNode.rightNode`.
+   1. If `node` is right of `parentNode`, add `{left: <leftHashString>}` to `peers`, where `leftHashString` is the hexadecimal string representation of the value at `parentNode.leftNode`.
+1. Return `peers`. 
+
+##### Validate Peers Array Map and Unsigned Beacon Signal
```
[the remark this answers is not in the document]
KDean-Dolphin: I agree that it's not a great name, but it's a map (keyed by the hash of the DID) of an array of peer nodes required to walk to the root of the SMT.

If we replace `peers` with something like `path`, then we could retitle it "Validate Paths Map and Unsigned Beacon Signal".

chapters/Beacons.md
```diff
@@ -0,0 +1,587 @@
...
+* `smtProofs` - required for services of `beaconType` "SMTBeacon", array of SMT proofs of inclusion or non-inclusion
+* `cas` - optional, one of:
+    * "ipfs"
+* `targetVersionId` - optional, DID document version ID required
+* `targetVersionTime` - optional, DID document version time required
+
+Process the ::Beacon Signals:: to reconstruct the DID document:
+
+1. If `cas` is defined and is not a valid value per above, raise InvalidParameter error.
+1. Set `sidecarDocumentsMap` to empty map.
```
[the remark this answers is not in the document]
KDean-Dolphin: Maybe `sidecarMap`?

I think a bigger problem is that we use "data" and "document" somewhat interchangeably. To me, "data" is the representation in memory, whereas "document" is what the data looks like when it's serialized as a string. That's part of what made the documentation for the map beacon difficult: we're creating a map (the data), serializing it (into a document), calculating the hash, and storing the document on a CAS. But, using the term "mapDocument" was causing some confusion, so I rewrote it, but I'm still not pleased with the result.

wip-abramson:
@KDean-Dolphin Lets merge this. And discuss the sidecar data in a separate issue.

PR #126, opened 2025-08-08
Add conformance section, drawing on DID core

wip-abramson:
This adapts language from DID Core and DID Resolution to define what it means to be conformant regarding this spec.

https://www.w3.org/TR/did/upcoming/#conformance

@peacekeeper would appreciate your review on this.

### DIFF
--- a/chapters/Conformance.md
+++ b/chapters/Conformance.md
@@ -6,14 +6,12 @@ The key words MAY, MUST, MUST NOT, RECOMMENDED, SHOULD, and SHOULD NOT in this d
 
 This document contains examples that contain JSON and JSON-LD content. Some of these examples contain characters that are invalid, such as inline comments (//) and the use of ellipsis (...) to denote information that adds little value to the example. Implementers are cautioned to remove this content if they desire to use the information as valid JSON or JSON-LD.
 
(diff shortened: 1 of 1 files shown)


peacekeeper (commented):
Here are two suggestions..

Also, I'd propose to remove the paragraphs about conforming producers and conforming consumers completely. This has to do with the abstract data model in DID Core at a time when we still differentiated between JSON and JSON-LD. And I don't think this makes much sense in a DID method spec.

peacekeeper:
Also, I'd propose to remove the paragraphs about conforming producers and conforming consumers completely. This has to do with the abstract data model in DID Core at a time when we still differentiated between JSON and JSON-LD. And I don't think this makes much sense in a DID method spec.

wip-abramson:
@peacekeeper hmm happy to get rid of conforming consumer that makes sense. But conforming producer (if that is what we want to call it) feels important.

This is some software that implements the Create Update and Deactivate algs from our spec.

peacekeeper:
> Can you propose some changes?

I'm sorry, I really don't know what else to say about it. My proposal would be to remove the text about producers and consumers, as I tried to explain. But if you want to keep it for some reason, I'm not strongly opposed to that either.

wip-abramson:
Okay I moved to registrar, merging

Issue #86, opened 2025-07-29
Align terminology for Beacons


jintekc:
The spec contains different terms for the same or similar concepts. A specific example of this is the Beacons sections.

## Problem

* BeaconAdvert vs. CohortAdvert vs. CohortSet
* Beacon Aggregation vs Aggregate Beacons
* CIDAggregate Beacon Advertisement vs. CIDAggregate Beacon Opt-in vs. Cohort Set

I realize that some of these variations are due to the nuances of human language (i.e. Beacon Aggregation describes an action vs. Aggregate Beacon describes an object). However, it starts to feel disconnected / loosely coupled making it challenging to mentally organize and connect the dots. We should work through the varying terms, their meanings, relations, etc. to come to a consolidate view of all the parts.

### Solution

Here's an example solution for one of the above problems, specifically aligning terms for CIDAggregate Beacons:

* CIDAggregate Beacon Advertisement => CIDAggregate Beacon Cohort Advertisement
* CIDAggregate Beacon Opt-in => CIDAggregate Beacon Cohort Opt-in
* Cohort Set => CIDAggregate Beacon Cohort Ready (I find the term "set" confusing as I immediately think of a mathematical "set" representing a unique group of items, a paradigm often used in programming languages).

wip-abramson:
@KDean-Dolphin is proposing migrating CIDAggregateBeacon to MapBeacon.

Please discuss...

wip-abramson:
@jintekc you should review this - https://github.com/dcdpr/did-btc1/compare/main...beacons-refactor

wip-abramson:
Hey @jintekc I have put together some Beacon terms for the terminology section that it would be good to get your eyes on.

https://docs.google.com/document/d/1DE8brHePdySsJPui0Zfn_8hpZlmhNcxB-lehAy_uw_M/edit?usp=sharing

This does not include the terms you mention about though I noticed. These do not feel like terms for the terminology section. Although I agree some alignment would be useful.

PR #146, opened 2025-08-10
Security Considerations Section Edits

jenniemeier:
Acronyms, caps, and links for the Security Considerations section.

@wip-abramson we capitalize the terms "Aggregators" and "Signals" in this section - should these remain capitalized or shift to lower-case?


wip-abramson:
Signals should be Beacon Signals I think.

And for Aggregators lets leave as is for now. We are a bit all over the place with our beacon terminology as @jintekc pointed out in #86.

Especially coordinators vs aggregators. We need to define the term for that role and use it throughout

PR #174, opened 2025-08-26
Add new Errors chapter

danpape:
I have added a new Errors chapter and tried to get all of the errors we mention in there. I have not yet gone through the whole spec to link the mentioned errors to the entries in this new chapter since I know you have a few big open pull requests. I figure once you get those merged then I'll go back and make sure everything is linked up correctly.

### DIFF
--- a/chapters/CRUD-Operations.md
+++ b/chapters/CRUD-Operations.md
@@ -331,7 +331,7 @@ It takes the following inputs:
   * `genesisBytes` - the originating ::Intermediate DID Document::; REQUIRED; bytes.
 * `initialDocument` - the ::Initial DID Document:: for the given identifier.
 
-It returns the following outputs or raises an error:
+It returns the following output or raises an error:
 
 * `initialDocument` - a valid ::Initial DID Document:: for the given identifier.
 
@@ -378,8 +378,8 @@ The steps are as follows:
 
 #### Resolve Target Document
 
-The Resolve Target Document algorithm resolves a DID document from an initial
-document by walking the Bitcoin blockchain to identify ::Beacon Signals:: that
+The Resolve Target Document algorithm resolves a DID document from an ::Initial DID Document::
+by walking the Bitcoin blockchain to identify ::Beacon Signals:: that
 announce ::BTC1 Updates:: applicable to the **did:btc1** identifier being
 resolved. 
 
@@ -412,7 +412,7 @@ The steps are as follows:
    `resolutionOptions.versionId`.
 1. Else if `resolutionOptions.versionTime` is not null, set `targetTime` to
    `resolutionOptions.versionTime`.
-1. Else set `targetTime` to the UNIX timestamp for now at the moment of execution.
+1. Else set `targetTime` to the UNIX timestamp for `now` at the moment of execution.
 1. Set `signalsMetadata` to `resolutionOptions.sidecarData.signalsMetadata`.
 1. Set `currentVersionId` to 1.
 1. If `currentVersionId` equals `targetVersionId` return `initialDocument`.
@@ -534,8 +534,8 @@ Note: It is recommended that you use a Bitcoin indexer and API to query the Bitc
 
 It takes the following inputs:
 
-* `contemporaryBlockheight` - The height of the block this function is looking for 
-   ::Beacon Signals:: in; REQUIRED; integer greater or equal to 0.
+* `contemporaryBlockheight` - The height of the block to look in for
+   ::Beacon Signals::; REQUIRED; integer greater or equal to 0.
 * `beacons` - An array of ::BTC1 Beacon:: services in the ::Contemporary DID document::; REQUIRED; array;
    Each Beacon is a structure with the following properties:
     * `id` - The id of the ::Beacon Service:: in the DID document; REQUIRED; string.
@@ -772,7 +772,7 @@ It returns the following output:
 The steps are as follows:
 
 1. Check that `sourceDocument.id` equals `btc1Identifier` else MUST raise
-   `invalidDidUpdate` error.
+   ::INVALID_DID_UPDATE:: error.
 1. Initialize `unsecuredBtc1Update` to an empty object.
 1. Set `unsecuredBtc1Update.@context` to the following list.
    `["https://w3id.org/zcap/v1", "https://w3id.org/security/data-integrity/v2", 
@@ -781,7 +781,7 @@ The steps are as follows:
 1. Set `targetDocument` to the result of applying the `documentPatch` to the
    `sourceDocument`, following the JSON Patch specification.
 1. Validate `targetDocument` is a conformant DID document, else MUST raise
-   `invalidDidUpdate` error.
+   ::INVALID_DID_UPDATE:: error.
 1. Set `sourceHashBytes` to the result of passing `sourceDocument` into
    the [JSON Canonicalization and Hash] algorithm.
 1. Set `unsecuredBtc1Update.sourceHash` to the base64 of `sourceHashBytes`.
--- a/chapters/Appendix.md
+++ b/chapters/Appendix.md
@@ -59,8 +59,6 @@ The function returns the retrieved `content` or null.
 
 ### Root did:btc1 Update Capabilities
 
-Note: Not sure if these algorithms should go here or in the appendix?
-
 #### Derive Root Capability from **did:btc1** Identifier
 
 This algorithm deterministically generates an [Authorization Capabilities for Linked Data](https://w3c-ccg.github.io/zcap-spec/) (ZCAP-LD) root capability from a given **did:btc1** identifier. Each root 
--- a/chapters/Beacons.md
+++ b/chapters/Beacons.md
@@ -80,7 +80,7 @@ Given:
     * "testnet3"
     * "testnet4"
     * "mutinynet"
-    * number
+    * number: 1, 2, 3 or 4
 * `serviceEndpoint` - required, a Bitcoin address represented as a URI
 * One and only one of, required:
     * `btc1Update` - the ::BTC1 Update:: whose SHA256 hash is the ::BTC1 Update Announcement:: to be broadcast in the ::Beacon Signal::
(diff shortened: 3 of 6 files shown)


chapters/Errors.md
```diff
@@ -0,0 +1,44 @@
+## Errors
+
+The algorithms described in this specification raise specific types of errors. Implementers might find it useful to convey these errors to other libraries or software systems. This section provides specific URLs and descriptions for the errors, such that an ecosystem implementing technologies described by this specification might interoperate more effectively when errors occur. Additionally, this specification uses some errors defined in Section 10 Errors of the [DID-RESOLUTION] specification. 
+
+Implementers SHOULD use [RFC9457] to encode the error data structure. If [RFC9457] is used:
```
wip-abramson, line 5:
```suggestion
Implementers SHOULD use [RFC9457] to encode the error data structure. If [RFC9457] is used:
```

Same here. This links wont work or render. They will just look like [RFC9457]

wip-abramson:
Hey Dan, this is a great start. I do not think we have to redefine the DID Resolution errors though. Just the errors that are specific to this specification.

We just state that we also use errors defined by DID Resolution.

Additionally, we need to identify all the areas in the spec that we talk about raising errors and do so in this new DID Resolution style. Everywhere we raise an error we should like to that error.

danpape:
Yep! Like I said above: "I have not yet gone through the whole spec to link the mentioned errors to the entries in this new chapter since I know you have a few big open pull requests. I figure once you get those merged then I'll go back and make sure everything is linked up correctly." It looks like some of them have already been merged, so I'll update on my end and push up some more changes. thanks

chapters/Errors.md
```diff
@@ -0,0 +1,44 @@
...
+
+- The type value of the error object MUST be a URL. Where the values listed in the section below do not define a URL, the values MUST be prepended with the URL https://btc1.dev/ns/error#.
+- The title value SHOULD provide a short but specific human-readable string for the error.
+- The detail value SHOULD provide a longer human-readable string for the error.
+
+BEACON_NOT_FOUND
+: No beacon service found in source document. See Section 4.3.3 Announce DID Update.
```
KDean-Dolphin, line 12: This should be a link rather than an explicit number.

chapters/Errors.md
```diff
@@ -0,0 +1,44 @@
...
+
+INCOMPLETE_SIDECAR_DATA
+: Missing sidecar data when processing an aggregate Beacon Signal.
```
KDean-Dolphin, line 15: Not all aggregate Beacon Signals have sidecar data, and sidecar data can appear in Singleton Beacons as well. The resolution algorithm allows for data to be in a mix of sidecar and CAS, so we should have a more generic "file not found" error.

chapters/Errors.md
```diff
@@ -0,0 +1,44 @@
...
+INVALID_DID
+: An invalid DID was detected during DID Resolution. See Section 4.4 DID Resolution Algorithm.
+
+INVALID_DID_DOCUMENT
+: An external initial document failed to validate as a conformant DID document according to the DID Core 1.1 specification.
+
+INVALID_DID_UPDATE
+: Problems when creating or applying a DID Update
+
+INVALID_HRP_VALUE
```
KDean-Dolphin, line 29: This should be folded into INVALID_DID. I don't see any value in having a separate type saying which part of the DID is invalid; that can be put into the `detail` property.

chapters/Errors.md
```diff
@@ -0,0 +1,44 @@
...
+: An invalid HRP value was found when decoding a **did:btc1** identifier. There are only two valid values: `k` for `key`, and `x` for `external`. See Section 3.2 did:btc1 Identifier Encoding.
+
+INVALID_SIDECAR_DATA
```
KDean-Dolphin, line 32: Given the resolution algorithm in the revised Beacons section, this will never occur, as files are indexed based on their hash, so a sidecar file that doesn't have a matching hash will manifest as "file not found".

chapters/Errors.md
```diff
@@ -0,0 +1,44 @@
...
+: A DID Update's contents differs from what was expected, and a hash of those contents after canonicalization did not match the hash given in the transaction or DID Update Bundle. 
+
+INVALID_UPDATE_PROOF
+: A DID Update failed to apply to a DID document because its `update.proof` was not verified. See Section 4.2.2.5 Apply DID Update.
+
+LATE_PUBLISHING
+: An error was found when processing the full history of DID Updates announced by all relevant Beacon Signals. See Section 6.1.1 Late Publishing.
```
KDean-Dolphin, line 39: Same comment re: link vs. explicit numbering.

chapters/Errors.md
```diff
@@ -0,0 +1,44 @@
...
+
+METHOD_NOT_SUPPORTED
+: An invalid DID method name was found when decoding a **did:btc1** identifier. See Section 3.3 did:btc1 Identifier Decoding.
+
```
KDean-Dolphin, line 43: Same comment re: link vs. explicit numbering.

KDean-Dolphin (dismissed):

chapters/Errors.md
wip-abramson: Kevin you confuse me again, by changing your term. What is a sidecar file here? The individual Sidecar Document required to validate a single Beacon Signal right?

I think this is an important error. We need to feedback to the client that there is missing sidecar data for a beacon signal with a specific ID.

In fact, I think MISSING_SIDECAR_DATA is important.

Reading this error message, I think it is still a valid error though. That is some BTC1 Update was announced by a Beacon Signal. But that BTC1 Update has an invalid signature.

chapters/Errors.md
wip-abramson:
```suggestion
: No beacon service found in source document. See Section [Announce DID Update].
```

chapters/Errors.md
wip-abramson: This should just be removed. No point defining errors that are already defined by DID Resolution

danpape: Well, maybe not re-define, but should we refer to the DID Resolution errors in the same way that the DID resolution spec refers to other external errors (like the last four errors here: https://w3c.github.io/did-resolution/#errors ) ?

wip-abramson:
> ```
> 1. If `idType` is not a valid value per above, raise ::https://www.w3.org/ns/did#INVALID_DID:: error.
> ```


I would do it like this:

If `idType` is not a valid value per above, raise an [INVALID_DID](https://www.w3.org/ns/did#INVALID_DID) error.

danpape:
> Note the url is https://w3c.github.io/did-resolution/#INVALID_DID

The DID Resolution spec says "https://www.w3.org/ns/did#."

https://w3c.github.io/did-resolution/#errors

???

rxgrant (approved):

chapters/CRUD-Operations.md
```diff
@@ -399,19 +399,20 @@ It takes the following inputs:
...
+It returns the following output or raises an error:
 
 * `targetDocument` - a DID Core conformant DID document after all updates have
    been found, validated and applied; object.
 
 The steps are as follows:
 
-1. If `resolutionOptions.versionId` is not null, set `targetVersionId` to
+1. If both `resolutionOptions.versionId` and `resolutionOptions.versionTime` are
+   not null then MUST raise an ::https://www.w3.org/ns/did#INVALID_DID:: error.
```
rxgrant, line 410: Let's add an INVALID_USAGE error, because the options are wrong, not the DID.

parasyte: Or `INVALID_OPTIONS` from DID Resolution: https://www.w3.org/TR/did-resolution/#errors

danpape: The DID Resolution spec defines `INVALID_OPTIONS` as an error for: "One or more of the options provided for DID Resolution or DID URL dereferencing are invalid." While the DID Resolution spec doesn't actually USE this error anywhere in its text, I assume it means that something in the `resolutionOptions` object is invalid. This doesn't really match the usage we want, which is reporting that some argument to one of our many algorithms/functions is invalid in some way. Our spec does define `INVALID_PARAMETER` as a somewhat generic error for "A parameter was invalid." and the details of exactly what was invalid would be provided within the `detail` part of the "Problem Details" structure. At this time, `INVALID_PARAMETER` is only used in a few places in the Beacons chapter, but I'll push up some changes which uses it in a few more places around the spec.

chapters/CRUD-Operations.md
```diff
@@ -163,7 +163,7 @@ The steps are as follows:
 1. Else If `identifierComponents.idType` value is "external", then set the
    `initialDocument` to the result of running [External Resolution] passing in
    the `identifier`, `identifierComponents` and `resolutionOptions` values.
-1. Else MUST raise `invalidHRPValue` error.
+1. Else MUST raise ::https://www.w3.org/ns/did#INVALID_DID:: error.
```
wip-abramson, line 166:
```suggestion
1. Else MUST raise ::https://www.w3.org/ns/did#INVALID_DID:: error.
```
Again, I am afraid it needs to just be a normal markdown link syntax

Do you know how to render the spec locally?

wip-abramson: "scripts": {
    "pandoc-spec": "pandoc-spec",
    "pandoc-spec-watch": "pandoc-spec --watch"
  },

chapters/CRUD-Operations.md
wip-abramson: Yep it is INVALID_OPTIONS for me. I raised an issue against DID resolution precisely because of this error state. INVALID_OPTIONS what what the WG came up with to address this

wip-abramson: > 1. If both `resolutionOptions.versionId` and `resolutionOptions.versionTime` are
   not null then MUST raise an ::https://www.w3.org/ns/did#INVALID_DID:: error.

This case here is saying that the resolutionOptions object provided is invalid.

PR #183, opened 2025-08-27
Urgent fix to address some terminology mismatches that have crept in

wip-abramson:


wip-abramson:
I am not sure I have caught all of them. But these are just the immediate ones that jumped out at me

Issue #178, opened 2025-08-26
Sidecar Data Terminology


wip-abramson:
As part of @KDean-Dolphin's PR on Beacons #100, it has become apparent that we need some betters terms from Sidecar Data.

Sidecar Data is the overarching name for data required to be provided via sidecar for the purposes of resolution.

Sidecar Data when provided in Resolution Options is a JSON document looking like this:

```json
"sidecarData": {
    "initialDocument": {
      "@context": [
        "https://www.w3.org/TR/did-1.1",
        "https://did-btc1/TBD/context"
      ],
      "id": "did:btc1:x1qyj23twdpn927d5ky2f5ulgmr9uudq2pd08gxy05fdjzxvfclzn2zazps8w",
      "controller": [
        "did:btc1:x1qyj23twdpn927d5ky2f5ulgmr9uudq2pd08gxy05fdjzxvfclzn2zazps8w"
      ],
      "verificationMethod": [
        {
          "id": "did:btc1:x1qyj23twdpn927d5ky2f5ulgmr9uudq2pd08gxy05fdjzxvfclzn2zazps8w#key-0",
          "type": "Multikey",
          "controller": "did:btc1:x1qyj23twdpn927d5ky2f5ulgmr9uudq2pd08gxy05fdjzxvfclzn2zazps8w",
          "publicKeyMultibase": "zQ3shYVjp5v4rRtg62akCYSeKKDdpvvTLRJxuac8rJEW9n72b"
        },
        {
          "id": "did:btc1:x1qyj23twdpn927d5ky2f5ulgmr9uudq2pd08gxy05fdjzxvfclzn2zazps8w#key-1",
          "type": "Multikey",
          "controller": "did:btc1:x1qyj23twdpn927d5ky2f5ulgmr9uudq2pd08gxy05fdjzxvfclzn2zazps8w",
          "publicKeyMultibase": "zQ3shhwxNrpzFpw2Xht1kdnupuUhgzVAbDiqjS9CZxtLg91jC"
        }
      ],
      "authentication": [
        "did:btc1:x1qyj23twdpn927d5ky2f5ulgmr9uudq2pd08gxy05fdjzxvfclzn2zazps8w#key-0"
      ],
      "capabilityInvocation": [
        "did:btc1:x1qyj23twdpn927d5ky2f5ulgmr9uudq2pd08gxy05fdjzxvfclzn2zazps8w#key-1"
      ],
      "capabilityDelegation": [
[...]

PR #193, opened 2025-08-28
Removed top of "Processing Signals"

KDean-Dolphin:
Removed top of "Processing Signals" to limit algorithm to individual signal types

### DIFF
--- a/chapters/Beacons.md
+++ b/chapters/Beacons.md
@@ -492,78 +492,38 @@ Spend the transaction:
 
 ### Processing Signals
 
+#### Process Singleton Beacon Signal
+
 Given:
 
-* `did` - required, the **did:btc1** identifier whose signals are to be processed
-* `sidecarDocuments` - required, array of documents required for resolution not stored on a ::CAS::,including:
-  * Initial DID document, if `did` was constructed with `idType` of "external" (`did` has the form "did:btc1:x1...")
-  * Map documents for ::BTC1 Beacons:: with services of `beaconType` "MapBeacon".
-  * ::BTC1 Updates:: for each ::BTC1 Update Announcement::
-* `smtProofs` - required for services of `beaconType` "SMTBeacon", array of SMT proofs of inclusion or non-inclusion
+* `hashBytes` - required, the hash bytes retrieved from the transaction
+* `sidecarDocumentsMap` - required, map to resolve documents not stored on a ::CAS::, keyed on document hash, including:
+    * ::BTC1 Updates:: for each ::BTC1 Update Announcement::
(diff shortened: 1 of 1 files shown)


danpape:
Looks good to me, thanks Kevin. I do see some issues with the passed arguments (in what is currently "7.2.2.3 Process Beacon Signals") to the algorithms in "chapters/Beacons.md". Do you want to fix those in this same PR?

wip-abramson:
Hmmm, yeah I am not sure the best way to address this. Some of this is about the rework to the new Sidecar Data structure #178. (We will no longer have maps from signalIds to sidecarData for that signal). I think this should be addressed in a separate PR.

But some are decisions we need to make, and should address in this PR I think. Specifically where do we retrieve the hashBytes (I actually wonder if we should define a term for this - Signal Bytes).

I think I like the way Kevin has done it, so that would mean updating the 7.2.2.3 Process Beacon Signals to first extract these bytes from the Beacon Signal, then pass it to the respective Process Beacon Signal algorithm. Although is this name then inaccurate? 

The only thing that gives me pause, is if in the future a new Beacon type constructs its Beacon Signal differently. So the signal bytes are not in the same location. But probably we can deal with that as and when it occurs.

wip-abramson:
Yep, sounds good. But what do you think about the interface into these Process Beacon Signals algorithms?

Should we keep it as you have it, taking signalBytes and the respective document maps? I think if we are already passing in the signalBytes, perhaps we should be passing in the sidecarData for those bytes not the maps

chapters/Beacons.md
```diff
@@ -492,79 +492,39 @@ Spend the transaction:
...
+
 Given:
 
-* `did` - required, the **did:btc1** identifier whose signals are to be processed
-* `sidecarDocuments` - required, array of documents required for resolution not stored on a ::CAS::,including:
-  * Initial DID document, if `did` was constructed with `idType` of "external" (`did` has the form "did:btc1:x1...")
-  * Map documents for ::BTC1 Beacons:: with services of `beaconType` "MapBeacon".
-  * ::BTC1 Updates:: for each ::BTC1 Update Announcement::
-* `smtProofs` - required for services of `beaconType` "SMTBeacon", array of SMT proofs of inclusion or non-inclusion
+* `hashBytes` - required, the hash bytes retrieved from the transaction
```
wip-abramson, line 499:
```suggestion
* `signalBytes` - required, the bytes retrieved from the last ::UTXO:: of the ::Beacon Signal::
```

wip-abramson:
Okay fair enough. Lets rebase and get this in, then get the PR that can align these two sections in ASAP

PR #204, opened 2025-08-31
Beacons Section Edits

jenniemeier:
Capitalization and links for the Beacons section.

TBD: In 6.2 Actors - we use the phrase aggregate BTC1 Beacon with only BTC1 Beacon linked. Should we add Aggregate BTC1 Beacon as a link option to Aggregate Beacon in terminology, or should they link separately to Aggregate and to BTC1 Beacon? 

TBD: In 6.6.2, we use the phrase SMT Beacon signal - how should we link for this one? Similar to the question above, and since there was no linking I added the SMT Beacon link for now, but that may not be sufficient.

### DIFF
--- a/chapters/Beacons.md
+++ b/chapters/Beacons.md
@@ -4,7 +4,7 @@
 
 A ::BCT1 Beacon:: is an abstract mechanism, identified by a Bitcoin address, that is included as a service in a DID document to indicate to resolvers that spends from the address, called ::Beacon Signals::, should be checked for ::BTC1 Update Announcements::.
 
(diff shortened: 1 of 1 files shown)


KDean-Dolphin:
There's a bit of weirdness in W3C when it comes to capitalization. In the DID specification, they (rightly, IMHO) talk about "Decentralized Identifiers" when referring to the artifact of the standard, and "decentralized identifiers" when referring to the generic concept. However, in the Verifiable Credentials specification, it's almost exclusively "verifiable credential" except when used in a section title.

I'm OK with capitalizing things that we define (e.g., Beacon Cohort) but not generic concepts (e.g., sparse Merkle tree, content-addressable storage). Other than that, I approve the changes.

PR #208, opened 2025-09-01
Add missing `controller` property in deterministic DID document generation

### LINKED ISSUE #206: Spec section 7.2.1.1 is missing a step to set the "controller" property
7.2.1.1 "Deterministically Generate Initial DID Document" describes the process to create a document deterministically. The documents in the test suite all have a `controller` property which is not mentioned in the Spec.

Omitting the `controller` property (per the spec) changes the `sourceHash` used in the test suite updates. In order to pass the test suite, the `controller` must be included in the deterministically generated DID document.

parasyte:
- Fixes #206


danpape (approved):
lgtm

parasyte:
I forced-pushed an identical commit to add a signing key.

PR #209, opened 2025-09-01
Fix formatting in Apply DID Update

parasyte:
- These two variables are strings.
- Use the same formatting as other sections where strings are used.

E.g.:

1. Set `idType` to "key".
1. Set `idType` to "external".


danpape (approved):
lgtm

rxgrant (approved):

PR #223, opened 2025-09-05
signalMetadata -> signalsMetadata

peacekeeper:
I think..?


parasyte (changes requested):
Line 886 references both `signalMetadata` and `signalsMetadata` as separate identifiers. One is a single unit of info, and the other is a list of units.

The flow chart also references these two identifiers by the same names: https://github.com/dcdpr/website-btc1-dev/blob/main/docs/public/diagrams/crud/update/sub3-announce-flowchart.mmd

It would be nice to pick a less ambiguous naming convention, though.

Issue #197, opened 2025-08-29
Add More Metadata to Spec Masthead


wip-abramson:
We should include in the spec its own URL and the URL for the github repository in the Masthead.

PR #224, opened 2025-09-08
Fix terminology section

wip-abramson:
This PR orders the terminology section alphabetically and removes duplicates.

It also adds a link to the spec into the spec itself in the metadata #197

### DIFF
--- a/chapters/Terminology.md
+++ b/chapters/Terminology.md
@@ -129,22 +129,6 @@ BTCR2 Update Announcements
 
 : ::BTCR2 Update Announcement::
 
-BTCR2 Update Announcement
-
-: ::BTCR2 Update Announcement::
-
-BTCR2 Update Announcements
-
-: ::BTCR2 Update Announcement::
-
-BTCR2 Update
-
-: A capability invocation secured using Data Integrity that invokes an authorization capability to update a specific **did:btcR2** DID document. This capability invocation Data Integrity proof secures the ::Unsecured BTCR2 Update:: document.
-
-BTCR2 Updates
-
-: ::BTCR2 Update::
-
 Content Addressable Storage
 
 : Content Addressable Storage (CAS) is a data storage system where content is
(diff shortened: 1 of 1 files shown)


chapters/Terminology.md
```diff
@@ -198,28 +211,38 @@ Merkle Trees
...
-: ::Sparse Merkle Tree::
+: Offline Creation refers to when a **did:btc1** identifier and corresponding
+  initial DID document are created without requiring network interactions.
 
-Spares Merkle Trees
+**did:btc1** supports offline creation in two modes:
 
-: ::Sparse Merkle Tree::
+* Key Pair Deterministic Creation; and
+* DID Document Initiated Creation.
```
parasyte, line 232: The CRUD chapter uses these subsection titles:

- From Deterministic Key Pair
- From External Intermediate DID Document

"Deterministic" aligns pretty well in both places. But the other mode omits the "External Intermediate" qualifier here. Consistency might be something worth addressing (or in another PR if changing the CRUD chapter).

chapters/Terminology.md
```diff
@@ -238,138 +261,104 @@ Schnorr
...
-**did:btc1** supports offline creation in two modes:
+: ::SMT Beacon::
 
-* Key Pair Deterministic Creation; and
-* DID Document Initiated Creation.
+Sparse Merkle Tree
 
-Sidecar
+: A ::Merkle Tree:: data structure where each data point included
+  at the leaf of the tree is indexed.
```
parasyte, line 304: This is a correct but incomplete description of a _sparse_ Merkle tree. I would add a definition for what makes it sparse:

```suggestion
: A ::Merkle Tree:: data structure where each data point included
  at the leaf of the tree is indexed. Empty leaves are discarded (zero values),
  making the tree sparse.
```

(Paraphrased from the alternative SMT appendix in #216.)

----

And it might not be appropriate to call attention to the optimizations (subtree collapsing etc.) in the terminology section. But if you wanted, you could also add something like this:

> Subtrees with a single path are collapsed to a single node, making the tree optimized.

PR #250, opened 2025-09-15
MapBeacon-service Update to did:btcr2

jenniemeier:
Updates references to did:btc1 to did:btcr2.


parasyte (approved):
We can make this change, but we'll also need to update any proofs that use this DID in an update patch.

wip-abramson (approved):

PR #270, opened 2025-10-10
added content to Conformance chapter

danpape:


parasyte:
I would personally prefer a "less is more" approach. Maybe two, three paragraphs max. Mostly citing normative references to specifications that define the conformance rules.

Of the RFCs we've been using as guidance, RFC 6902 only has a single paragraph for this purpose:

https://www.rfc-editor.org/rfc/rfc6902.html#section-2

> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
> "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
> document are to be interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) [[RFC2119](https://www.rfc-editor.org/rfc/rfc2119)].

Meanwhile, normative references like, "JavaScript Object Notation (JSON) [[RFC4627](https://www.rfc-editor.org/rfc/rfc4627)]" and "JSON-Pointer [[RFC6901](https://www.rfc-editor.org/rfc/rfc6901)]" are used in various parts of the specification as necessary.

PR #286, opened 2026-01-16
Intro Section Edits

jenniemeier:
Links, bolding, and very small verbiage adjustments for the Introduction and Motivation Section.

PR #290, opened 2026-01-17
BTCR2 Update Data Distribution Section Edits

jenniemeier:
Added missing links, slight grammar formatting, unwrapping of long lines.

### DIFF
--- a/src/update-data-distribution.md
+++ b/src/update-data-distribution.md
@@ -5,40 +5,74 @@
 
 # BTCR2 Update Data Distribution
 
-In many cases, additional data is required for resolution. In those cases, the resolver can receive data via two primary means from a DID controller, both secured by cryptographic hash:
+In many cases, additional data is required for resolution. In those cases, 
+the resolver can receive data via two primary means from a DID controller, both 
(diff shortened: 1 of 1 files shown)


jintekc (approved):

parasyte (dismissed):
> ... unwrapping of long lines.

Unless I am terribly mistaken, this change _wraps_ long lines at a specific character column into multiple short lines. Unwrapping would be the opposite, combining multiple short lines into one long line.

I strongly prefer **UNwrapping**, which is opposite of what this PR is doing. The diff is impossible to read at a glance with hard-wrapped lines. And it is not what I had in mind for https://github.com/dcdpr/did-btcr2/pull/286#discussion_r2699750505

For that reason, I'm going to say no to this change. I might be vetoed, but this is a big readability issue for change tracking.

danpape:
It appears that some of our files have nice long lines: appendix.md, conformance.md, and update-data-distribution.md, but the rest have already been reflowed to short lines (or maybe they were like like that to begin with?).

danpape (approved):

PR #292, opened 2026-01-19
Aggregate Beacons Section Edits

jenniemeier:
Added / updated links, minor changes to bullet list punctuation.

A couple notes on this section:

- In areas where CAS Beacon Signals and SMT Beacon Signals are used, I updated the terminology linking from the more general Beacon Signals to CAS Beacon or SMT Beacon as both of these terms contain details about their specific Beacon Signals in their definitions. However, given the number of times these terms appear in the spec, I think it would be a good idea to consider adding the terms CAS Beacon Signal and SMT Beacon Signal to the terminology. I'll open this in an Issue for discussion. 
- We are inconsistent in the way we do our bullet point list punctuation, and while this is easy to fix in some of the simple lists, it becomes less clear when sub-bullets and sub-sub-bullets are involved, as well as when a bullet contains several sentences. I'll open an Issue for discussion on this as well.

### DIFF
--- a/src/beacons/aggregate-beacons.md
+++ b/src/beacons/aggregate-beacons.md
@@ -5,32 +5,32 @@
 
 # Aggregate Beacons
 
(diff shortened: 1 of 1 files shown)


danpape (approved):
Starting to waver, I think I do like the semi-colon formatted lists.  I guess we should have a separate PR later to make sure everything is consistent everywhere.

jintekc (approved):

PR #297, opened 2026-01-30
Fix typo on conformance.md

jintekc:
Misspelling of DID-RESOLUTION


parasyte:
I thought the mdbook build action that we already have would do it. https://github.com/dcdpr/did-btcr2/blob/f2c3fb67c7670e9a225a4407271f64f04b0e24a8/.github/workflows/deploy.yml#L29

The plugin is installed. I guess it just has a bug where it doesn't do validation. The only thing I can find right now is a way to [enable debug logging](https://francisco-perez-sorrosal.github.io/mdbook-bib/dev.html#debug). Maybe that could help identify the problem?

PR #316, opened 2026-05-08
Mention that CAS Announcement can be retrieved from CAS

peacekeeper:


danpape:
This seems good to me, but I guess I would ask: should we say something like "..., attempt to retrieve it from [CAS]. Raise a [MISSING_UPDATE_DATA] error if the update is not available." But that would require we modify the text in earlier section "Process Beacon Signals" after the bullet point "Build a tuple with:"

wip-abramson:
SMT proofs cannot be retrieved from CAS. 

At least not without additional information. There is no hash that identifies the proof in the Beacon Signal.

peacekeeper:
> SMT proofs cannot be retrieved from CAS.

Right, thanks.. I had a feeling something was wrong here. I force-pushed a commit to remove this part.

jintekc (approved):

danpape (approved):

danpape:
Hey @peacekeeper we can't merge your commit as you don't have commit signing set up. Is that something you think you can take care of?

peacekeeper:
@danpape Sorry it took me a while. I signed my commit now.

PR #319, opened 2026-06-16
Align optimized SMT appendix with hashed-zero algorithm and fix example encoding 

### LINKED ISSUE #317: Conflicting SMT Proof Verification Algorithms
The spec contains two conflicting SMT proof verification algorithms.

The algorithm at the end of [appendix/optimized-smt](https://dcdpr.github.io/did-btcr2/appendix/optimized-smt.html) skips hashing if the collpased bit = 1.

> With an [SMT Proof](https://dcdpr.github.io/did-btcr2/terminology.html#smt-proof), the verifier has everything necessary to verify that the [BTCR2 Update](https://dcdpr.github.io/did-btcr2/terminology.html#btcr2-update) is included or not. Assuming that the verifier can match the root hash in the [Beacon Signal](https://dcdpr.github.io/did-btcr2/terminology.html#beacon-signal) to `proof.id` and the [BTCR2 Update](https://dcdpr.github.io/did-btcr2/terminology.html#btcr2-update) to `proof.updateId`, the proof is verified as follows:
>
> ```javascript
> index = int(hash(did)) // 1101
> 
> candidateHash = hash(hash(proof.nonce) + proof.updateId)
> 
> // First collapsed bit from right is 1, so index bit doesn't apply.
> // Skip first index bit.
> // Candidate hash is unchanged.
> 
> // Next collapsed bit from right is 0, so index bit applies.
> // Next index bit from right is 0.
> // Candidate hash goes to the left against the next listed hash.
> candidateHash = hash(candidateHash + "Hash 1110")
> 
> // Next collapsed bit from right is 0, so index bit applies.
> // Next index bit from right is 1.
> // Candidate hash goes to the right against the next listed hash.
> candidateHash = hash("Hash 1001" + candidateHash)
> 
> // Next collapsed bit from right is 0, so index bit applies.
> // Next index bit from right is 1.
> // Candidate hash goes to the right against the next listed hash.
> candidateHash = hash("Hash 0" + candidateHash)
> 
> // Hashes exhausted.
> // Candidate hash must equal root hash.
> assert(candidateHash === proof.id)
> ```

The [SMT Proof Verification](https://dcdpr.github.io/did-btcr2/algorithms.html#smt-proof-verification) algorithm on the Algorithms page hashes the candidateHash with has pre-computed hash of two zeros pulled from a hashed-zero cache.

> The tree is “optimized” by collapsing empty nodes. (See [Appendix: Optimized Sparse Merkle Tree Implementation](https://dcdpr.github.io/did-btcr2/appendix.html#smt-implementation) for definition of “optimized”.) Each bit within `proof.collapsed` informs the algorithm whether to take the next sibling hash from `proof.hashes` (0) or use a hashed zero from the current height of the tree (1).

```rust
let candidateHash = hash(concat(hash(proof.nonce), proof.updateId))
let index = hash(did)
for n in 0..=255 {
  let i = 255 - n;

  let siblingHash = if proof.collapsed[i] == 1 {
    cachedZero[n]
  } else {
    proof.hashes.pop_front()
  };

  if index[i] == 1 {
    candidateHash = hash(concat(siblingHash, candidateHash));
  } else {
    candidateHash = hash(concat(candidateHash, siblingHash));
  }
}

return candidateHash == proof.id;
```

These are incompatible and change how proofs are produced and verified.

### LINKED ISSUE #318: Conflicting SMT Proof Data Structure Defs
The [SMT Proof (data structure)](https://dcdpr.github.io/did-btcr2/data-structures.html#sidecar-smt-proof-example-panel-show) defines fields as "MUST be "base64url" [[RFC4648](https://dcdpr.github.io/did-btcr2/bibliography.html#RFC4648)] encoded without padding." However, the actual examples say "Hexidecimal of ..."

```json
{
  "id": "<< Hexadecimal of Root Hash >>",
  "nonce": "<< Hexadecimal of Nonce 1101 >>",
  "updateId": "<< Hexadecimal of hash(Data Block 1101) >>",
  "collapsed": "<< Hexadecimal of 0001 >>",
  "hashes": [
      "<< Hexadecimal of Hash 1110 >>",
      "<< Hexadecimal of Hash 1001 >>",
      "<< Hexadecimal of Hash 0 >>"
  ]
}
``` 

The code examples should align to the listed def requirements of base64url no padding.

jintekc:
Closes #317 
Closes #318

### DIFF
--- a/src/data-structures.md
+++ b/src/data-structures.md
@@ -253,8 +253,8 @@ SHA-256 hashes {{#cite SHA256}} (`id`, `updateId`, `hashes`) MUST be `"base64url
 - `id`: SHA-256 hash of the root node.
 - `nonce`: OPTIONAL 256-bit nonce generated for each update. MUST be encoded as a string using `"base64url"` {{#cite RFC4648}} encoding without padding.
 - `updateId`: The OPTIONAL [BTCR2 Signed Update (data structure)] hashed with the [JSON Document Hashing] algorithm.
-- `collapsed`: Bitmap of zero nodes within the path (see: [collapsed leaves](https://github.com/hoytech/quadrable#collapsed-leaves)).
-- `hashes`: Array of SHA-256 hashes representing the sibling [SMT] nodes from the leaf, containing the SHA-256 hash of the [BTCR2 Signed Update] or the "zero identity", to the root.
+- `collapsed`: Bitmap of zero nodes within the path (see: [collapsed leaves](https://github.com/hoytech/quadrable#collapsed-leaves)). MUST be `"base64url"` {{#cite RFC4648}} encoded without padding.
+- `hashes`: Array of SHA-256 hashes representing the sibling [SMT] nodes from the leaf, containing the SHA-256 hash of the [BTCR2 Signed Update] or the "zero identity", to the root. Each hash MUST be `"base64url"` {{#cite RFC4648}} encoded without padding.
 
 
 {% set hide_text = `` %}
--- a/src/algorithms.md
+++ b/src/algorithms.md
@@ -128,6 +128,8 @@ Example output:
 
 To verify the inclusion or non-inclusion of a DID in the [SMT Proof], perform the following steps:
 
+Throughout this section, `hash()` denotes SHA-256 {{#cite SHA256}} over a byte array, and `concat()` (equivalently, the `+` operator) concatenates byte arrays, not strings. The `base64url` {{#cite RFC4648}} encoded fields of an [SMT Proof (data structure)] (`id`, `nonce`, `updateId`, `collapsed`, and the entries of `hashes`) MUST be decoded to their raw bytes before being used in any of these operations.
+
 Construct a hashed-zero cache. Two zero leaves are hashed together by concatenating the value `0` with itself for the bottom level, and then hashed again with its next sibling, also itself, at each level up the tree. Each computed hash value can be cached in an array.
 
 {% set hide_text = `` %}
--- a/src/appendix/optimized-smt.md
+++ b/src/appendix/optimized-smt.md
@@ -121,15 +121,14 @@ To use Merkle trees to signal commitments in [BTCR2 Beacons][BTCR2 Beacon]:
 * The value stored at a leaf node is the hash of a 256-bit nonce, concatenated with the hash of the [BTCR2 Update] (the [BTCR2 Update Announcement]) if available, then the resulting stream is hashed again. I.e., `value = hash(hash(nonce) + hash(btcr2Update))` if there is a [BTCR2 Update] or `value = hash(hash(nonce))` if there is not.
     * Provided that it is unique per DID and per signal, the use of a nonce ensures that updates and non-updates are indistinguishable to outside parties (aggregators, other DID controllers, verifiers) unless explicitly informed by the DID controller.
     * The hashing of the nonce ensures that verifiers with limited input validation deal only with a 256-bit result.
-* A node with two empty children is itself empty.
-* The value of a node with one empty child and one non-empty child is the value of the non-empty child.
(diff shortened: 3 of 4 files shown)


parasyte:
One thing that might be useful to avoid some confusion is specifying that the concatenation operator (`+`) and `concat()` function takes a byte array, not a string. And that the encoded hashes in SMT proofs must be decoded before operating on them. If this is already specified somewhere and I missed it, please point it out.

I can imagine a scenario where the word "concatenation" is interpreted as "string concatenation" instead of "byte array concatenation".

Also, the SMT Proof Verification algorithm description and pseudocode currently do not work with "non-updates" (i.e., with a proof of non-inclusion). It's a simple fix; make the concatenation of hashed-nonce and update hash optional (while maintaining the outer hash - very important). The `updateId` property on the SMT Proof data structure is already optional, but the algorithm and pseudocode do not acknowledge that.

----

Still reviewing the changes. This is just a thought that came up while going over it.

src/data-structures.md
```diff
@@ -253,7 +253,7 @@ SHA-256 hashes {{#cite SHA256}} (`id`, `updateId`, `hashes`) MUST be `"base64url
 - `id`: SHA-256 hash of the root node.
 - `nonce`: OPTIONAL 256-bit nonce generated for each update. MUST be encoded as a string using `"base64url"` {{#cite RFC4648}} encoding without padding.
 - `updateId`: The OPTIONAL [BTCR2 Signed Update (data structure)] hashed with the [JSON Document Hashing] algorithm.
-- `collapsed`: Bitmap of zero nodes within the path (see: [collapsed leaves](https://github.com/hoytech/quadrable#collapsed-leaves)).
+- `collapsed`: Bitmap of zero nodes within the path (see: [collapsed leaves](https://github.com/hoytech/quadrable#collapsed-leaves)). MUST be `"base64url"` {{#cite RFC4648}} encoded without padding.
 - `hashes`: Array of SHA-256 hashes representing the sibling [SMT] nodes from the leaf, containing the SHA-256 hash of the [BTCR2 Signed Update] or the "zero identity", to the root.
```
parasyte: Nitpick: This could also specify that each hash MUST be `"base64url"` encoded without padding.

parasyte (dismissed):
Looks great! I found only one minor suggestion.

The new diagram reads better (IMHO) with every level of the tree covered and each node receiving a meaningful identifier. I'll create a separate PR for fixing the unrelated SMT Proof Verification algorithm issue I noticed.

jintekc:
> Looks great! I found only one minor suggestion.
> 
> The new diagram reads better (IMHO) with every level of the tree covered and each node receiving a meaningful identifier. I'll create a separate PR for fixing the unrelated SMT Proof Verification algorithm issue I noticed.

Thanks for the prompt review! Addressed here:

1. added a note that `hash()/concat()/+` operate on byte arrays (not strings) and that the base64url proof fields must be decoded to raw bytes first
2. the hashes bullet now states each hash MUST be base64url without padding. Left the non-update (proof-of-non-inclusion) algorithm fix for your separate PR, as you suggested.

parasyte (dismissed):
The `+` operator is also used in the Optimized SMT chapter. But this works for me.

danpape:
I also approve, looks great! Like Jay mentioned though, maybe you can also add the "Throughout this section, `hash()` denotes SHA-256..." paragraph from src/algorithms.md to src/appendix/optimized-smt.md?

jintekc:
@danpape @parasyte - pushed one follow-up commit addressing the review: added the `hash()/+` byte-array and base64url-decoding note to `src/appendix/optimized-smt.md`, and linked the appendix to the normative `[SMT Proof Verification]` algorithm. No algorithm changes. Needs two more approvals.

danpape (approved):

parasyte (approved):

PR #320, opened 2026-06-16
Fix proof of non-inclusion in SMT Proof Verification algorithm

parasyte:
This was discussed in https://github.com/dcdpr/did-btcr2/pull/319#issuecomment-4723580108

Adds support for non-inclusion proofs to the algorithm description and pseudocode. Fixes some minor syntactic errors (missing semicolons) and slightly improves readability with whitespace.

The `cachedZero` pseudocode was also updated to use `n` as the index variable name to better match the longer pseudocode.

And today I learned that mdBook's `smart-punctuation` option allows rendering em dashes with triple-hyphens. https://rust-lang.github.io/mdBook/format/markdown.html#smart-punctuation Em dashes looked better than parentheses and commas.

### DIFF
--- a/src/algorithms.md
+++ b/src/algorithms.md
@@ -136,9 +136,10 @@ Construct a hashed-zero cache. Two zero leaves are hashed together by concatenat
 ~~~rust
 let cachedZero = [];
 let z = 0;
-for i in 0..=255 {
+
+for n in 0..=255 {
   z = hash(concat(z, z));
-  cachedZero[i] = z;
+  cachedZero[n] = z;
 }
 ~~~
 ` %}
@@ -153,7 +154,7 @@ hide_label="Hide"
 ) }}
 
 
-The [SMT Proof (data structure)] is verified by walking the tree starting from the leaf node value, given by `hash(hash(proof.nonce) + proof.updateId)`, to the root `proof.id`. Walking the tree means hashing sibling hashes from `proof.hashes` concatenated with a candidate hash to construct each node value. Each bit within the leaf node index (given by `hash(did)`) informs the algorithm which side of the concatenation operation the sibling hash belongs on: 0 for left, 1 for right.
+The [SMT Proof (data structure)] is verified by walking the tree starting from the leaf node value --- given by `hash(hash(proof.nonce) + proof.updateId)` for update inclusion or `hash(hash(proof.nonce))` for non-inclusion --- to the root `proof.id`. Walking the tree means hashing sibling hashes from `proof.hashes` concatenated with a candidate hash to construct each node value. Each bit within the leaf node index (given by `hash(did)`) informs the algorithm which side of the concatenation operation the sibling hash belongs on: 0 for left, 1 for right.
 
 The tree is "optimized" by collapsing empty nodes. (See [Appendix: Optimized Sparse Merkle Tree Implementation] for definition of "optimized".) Each bit within `proof.collapsed` informs the algorithm whether to take the next sibling hash from `proof.hashes` (0) or use a hashed zero from the current height of the tree (1).
 
@@ -163,18 +164,24 @@ The last step is asserting that the computed candidate hash is equivalent to `pr
 {% set pseudocode_smt_proof_verification =
 `
 ~~~rust
-let candidateHash = hash(concat(hash(proof.nonce), proof.updateId))
-let index = hash(did)
+let candidateHash = if let Some(updateId) = proof.updateId {
+  hash(concat(hash(proof.nonce), updateId))
+} else {
+  hash(hash(proof.nonce))
+};
+
+let index = hash(did);
+
 for n in 0..=255 {
   let i = 255 - n;
 
-  let siblingHash = if proof.collapsed[i] == 1 {
+  let siblingHash = if proof.collapsed.bitAt(i) == 1 {
     cachedZero[n]
   } else {
     proof.hashes.pop_front()
   };
 
-  if index[i] == 1 {
+  if index.bitAt(i) == 1 {
     candidateHash = hash(concat(siblingHash, candidateHash));
   } else {
     candidateHash = hash(concat(candidateHash, siblingHash));


parasyte:
Another change I made that I forgot to note: The bit-indexing was replaced with a `.bitAt()` method. I can't explain why I used `camelCase`, but there it is. I can change it to `snake_case` if preferred. I think the method's intention is clear without additional commentary, but maybe it needs some extra exposition.

src/algorithms.md
```diff
@@ -163,18 +164,24 @@ The last step is asserting that the computed candidate hash is equivalent to `pr
...
+  hash(hash(proof.nonce))
+};
+
+let index = hash(did);
+
 for n in 0..=255 {
   let i = 255 - n;
 
-  let siblingHash = if proof.collapsed[i] == 1 {
+  let siblingHash = if proof.collapsed.bitAt(i) == 1 {
```
jintekc, line 178: Since `bitAt` is new, should we define the bit order explicitly? Maybe add one sentence so implementers don't guess: "bitAt(0) is the most-significant bit and bitAt(255) is the least significant bit." Might be "belt-and-suspenders" but figured I would just note it.

jintekc:
on a separate and unrelated note. I think `nonce` should be `REQUIRED`, not `OPTIONAL` in `data-structures.md`. Both leaf-value branches compute `hash(hash(proof.nonce))`, so every proof needs the nonce. Maybe we could sneak that into this PR as well.

jintekc (approved):

src/algorithms.md
danpape: Wait, what is the type of `proof.collapsed`? I guess this is just pseudocode that looks a lot like rust, but I'll note that in the data structure chapter, under SMT Proof, we describe it as 

`collapsed`: Bitmap of zero nodes within the path

and in the example JSON provided, it of course is just a string:

"collapsed": "<< Hexadecimal of 0001 >>",

so I guess it is left to the implementer to figure out how to ingest that string as a Bitmap-like type or class and hopefully have access to some kind of `bitAt` function. 

In any case, if WE think our pseudo function acts like "bitAt(0) is the most-significant bit..." then we should probably mention it here--maybe as a comment in this pseudocode?

danpape:
> on a separate and unrelated note. I think `nonce` should be `REQUIRED`, not `OPTIONAL` ...

Hmm, yeah, we did write it like

nonce: OPTIONAL 256-bit nonce generated for each update. MUST be encoded as a string ...

I sure don't remember whatever discussion we had last year about why this might be optional or not, but it does seem like we kind of require it now.

src/algorithms.md
jintekc: So in my PR, I added a clarifying statement that the base64urlnopad strings should be decoded to their raw bytes before being used in ops. once this lands, it'll fit nicely into the rust pseudocode above

https://github.com/dcdpr/did-btcr2/pull/319/changes#diff-0784dc6d7a2f21b84698991b0273d7a28a24a12eff638ba394aa0a7d51ef3dccR131

> Throughout this section, `hash()` denotes SHA-256 {{#cite SHA256}} over a byte array, and `concat()` (equivalently, the `+` operator) concatenates byte arrays, not strings. The `base64url` {{#cite RFC4648}} encoded fields of an [SMT Proof (data structure)] (`id`, `nonce`, `updateId`, `collapsed`, and the entries of `hashes`) MUST be decoded to their raw bytes before being used in any of these operations.

parasyte:
`nonce` was made optional because it bloats proofs when the additional privacy isn't needed. A proof of non-inclusion is slightly smaller without a nonce, and as the number of aggregate update rounds increases, it can eventually create a huge number of proofs, almost all of them non-inclusions.

@rxgrant had some thoughts on this, specifically.

I don't have a problem with it either way. The `nonce` is not required for security. It increases privacy by making all proofs (including non-inclusion) indistinguishable from proofs of inclusion. Whether that privacy is useful is an end-user decision.

src/algorithms.md
parasyte: This is Rust-shaped pseudocode, but the type of `proof.collapsed` is intended to be roughly `[u8; 32]`; the raw bytes of a SHA-256 hash or a newtype around it.

`bitAt()` is intended to be a bit-index (counting from left-to-right), implementing the "each bit" check described in:

> Each bit within `proof.collapsed` informs the algorithm whether to take the next sibling hash from `proof.hashes` (0) or use a hashed zero from the current height of the tree (1).

Previously, the pseudocode used an array-like bit-index operator `[]`, which isn't real Rust either. :) Both are approximately obvious to what they do, at least to me. I was hoping that naming the operator instead of using a sigil-pair made it more apparent. If `bitAt()` doesn't increase clarity, maybe a different name will, in place of a prose description of its purpose: `bit_index()` or `bit_index_from_left_to_right()`?

parasyte: Oh, and another thought I had: Algorithms are supposed to operate on data types deserialized from JSON, and never on JSON directly. JSON is just a transmission layer encoding for *Data Structures*. It has some of its own encoding rules, such as how to encode SHA-256 hashes or other 256-bit integers (like the `collapsed` bit mask and `nonce`).

Any algorithm that isn't directly implementing the specified *Data Structures* should be considered as operating on a very different data model that more closely resembles the C abstract machine (but not precisely -- there is no need for an ABI specification). The reason for this split is so that the spec does not force everything to be JSON-shaped throughout all of the algorithms.

Admittedly, there isn't much of a data model description. The Algorithms chapter strongly implies that byte arrays are used, which isn't a native JSON type. I think that's about as close as it gets at present.

jintekc: > This is Rust-shaped pseudocode, but the type of `proof.collapsed` is intended to be roughly `[u8; 32]`; the raw bytes of a SHA-256 hash or a newtype around it.
> 
> `bitAt()` is intended to be a bit-index (counting from left-to-right), implementing the "each bit" check described in:
> 
> > Each bit within `proof.collapsed` informs the algorithm whether to take the next sibling hash from `proof.hashes` (0) or use a hashed zero from the current height of the tree (1).
> 
> Previously, the pseudocode used an array-like bit-index operator `[]`, which isn't real Rust either. :) Both are approximately obvious to what they do, at least to me. I was hoping that naming the operator instead of using a sigil-pair made it more apparent. If `bitAt()` doesn't increase clarity, maybe a different name will, in place of a prose description of its purpose: `bit_index()` or `bit_index_from_left_to_right()`?

yeah fair enough. I'm okay with leaving as-is.

danpape (approved):

PR #337, opened 2026-07-27
Only increment current_version_id when an update is applied

### LINKED ISSUE #321: Duplicate updates shouldn't bump `current_version_id`
## Problem

In [Process updates Array](https://dcdpr.github.io/did-btcr2/operations/resolve.html#process-updates), step 5 "Increment `current_version_id`" runs for *every* tuple, including ones that got sent to [Confirm Duplicate Update](https://dcdpr.github.io/did-btcr2/operations/resolve.html#confirm-duplicate-update). Confirm Duplicate Update only compares, it never appends to `update_hash_history`, so bumping the counter on a duplicate pushes `current_version_id` one ahead of the history for each duplicate seen.

Two things break

1. **Inflated version.** `didDocumentMetadata.versionId` comes back one too high per trailing duplicate.
2. **A valid history stops resolving.** If a real update follows a duplicate, the inflated counter makes it look like `targetVersionId <= current_version_id`, so it gets misrouted into Confirm Duplicate Update. That reads `update_hash_history[targetVersionId - 2]`, which is now out of range, and you get a false `LATE_PUBLISHING`. So a history like `[v2, v2-again, v3]` fails even though every update is well formed.

## Solution

Only bump the counter when you actually apply an update; move the increment out of "Process updates Array" and into "Apply update":

**Apply update** (add the increment as the last step):
> Create unsigned_update by removing the proof property from update, hash it with the JSON Document Hashing algorithm, and append the hash to `update_hash_history`. Increment `current_version_id`.

Confirm Duplicate Update stays exactly as it is. This keeps `current_version_id` and the history in lockstep (`current_version_id == update_hash_history.length + 1`).

## Related

- #303 covers the loop/return ambiguity in this same section. Fixing both would fully pin down the loop.

jintekc:
Moves the `current_version_id` increment out of the "Process updates Array" loop and into "Apply update", so duplicates confirmed via "Confirm Duplicate Update" no longer advance the counter. This keeps `current_version_id` in lockstep with `update_hash_history` (fixes the inflated `versionId` and the false `LATE_PUBLISHING` on histories like [v2, v2-again, v3]).
  
Also adds the guard discussed in the issue thread: "Confirm Duplicate Update" now raises `INVALID_DID_UPDATE` when `targetVersionId < 2`, so a non-conforming update can't reach the history lookup with a negative index and all resolvers reject it the same way.
  
Fixes #321

### DIFF
--- a/src/operations/resolve.md
+++ b/src/operations/resolve.md
@@ -189,7 +189,9 @@ Verify that `current_document` conforms to DID Core v1.1 {{#cite DID-CORE}} and 
 
 Hash the patched `current_document` with the [JSON Document Hashing] algorithm. Raise an [`INVALID_DID_UPDATE`] error if the result does not match the decoded `update.targetHash`.
 
-Create `unsigned_update` by removing the `proof` property from `update`, hash it with the [JSON Document Hashing] algorithm, and append the hash to `update_hash_history`. Increment `current_version_id`.
+Create `unsigned_update` by removing the `proof` property from `update`, hash it with the [JSON Document Hashing] algorithm, and append the hash to `update_hash_history`. 
+
+Increment `current_version_id`.
 
 
 ### Check `update.proof` { #check-update-proof }
--- a/src/operations/resolve.md
+++ b/src/operations/resolve.md
@@ -152,9 +152,8 @@ Otherwise:
 2. Set `block_confirmations` to the tuple's block confirmations.
 3. If `resolutionOptions.versionTime` is provided and the tuple's block time is more recent, resolve `current_document` as `didDocument`.
 4. Set `update` to the tuple's [BTCR2 Signed Update (data structure)] and [check `update.targetVersionId`](#check-update-version).
-5. Increment `current_version_id`.
-6. If `current_version_id` is greater than or equal to the integer form of `resolutionOptions.versionId`, resolve `current_document` as `didDocument`.
-7. If `current_document.deactivated` is `true`, resolve `current_document` as `didDocument`.
+5. If `current_version_id` is greater than or equal to the integer form of `resolutionOptions.versionId`, resolve `current_document` as `didDocument`.
+6. If `current_document.deactivated` is `true`, resolve `current_document` as `didDocument`.
 
 
 ### Check `update.targetVersionId` { #check-update-version }
@@ -188,7 +187,7 @@ Verify that `current_document` conforms to DID Core v1.1 {{#cite DID-CORE}} and 
 
 Hash the patched `current_document` with the [JSON Document Hashing] algorithm. Raise an [`INVALID_DID_UPDATE`] error if the result does not match the decoded `update.targetHash`.
 
-Create `unsigned_update` by removing the `proof` property from `update`, hash it with the [JSON Document Hashing] algorithm, and append the hash to `update_hash_history`.
+Create `unsigned_update` by removing the `proof` property from `update`, hash it with the [JSON Document Hashing] algorithm, and append the hash to `update_hash_history`. Increment `current_version_id`.
 
 
 ### Check `update.proof` { #check-update-proof }
--- a/src/operations/resolve.md
+++ b/src/operations/resolve.md
@@ -172,6 +172,8 @@ Compare `update.targetVersionId` to `current_version_id`. Only one of three poss
 
 This step confirms that an update with a lower-than-expected `targetVersionId` is a true duplicate.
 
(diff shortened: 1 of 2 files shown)


src/operations/resolve.md
```diff
@@ -188,7 +189,7 @@ Verify that `current_document` conforms to DID Core v1.1 {{#cite DID-CORE}} and
 
 Hash the patched `current_document` with the [JSON Document Hashing] algorithm. Raise an [`INVALID_DID_UPDATE`] error if the result does not match the decoded `update.targetHash`.
 
-Create `unsigned_update` by removing the `proof` property from `update`, hash it with the [JSON Document Hashing] algorithm, and append the hash to `update_hash_history`.
+Create `unsigned_update` by removing the `proof` property from `update`, hash it with the [JSON Document Hashing] algorithm, and append the hash to `update_hash_history`. Increment `current_version_id`.
```
danpape: Seems like the increment could be its own separate "step"?

```suggestion
Create `unsigned_update` by removing the `proof` property from `update`, hash it with the [JSON Document Hashing] algorithm, and append the hash to `update_hash_history`. 

Increment `current_version_id`.
```

danpape (dismissed):
Looks good.

parasyte:
I'm confused. "Confirm Duplicate Update" confirms a duplicate by raising an error. When the error is raised, processing the `updates` array (and everything else in the resolver, really) stops. That will never reach the increment.

I agree with the new "if `update.targetVersionId` is less than `2`" condition (we discussed this in #321). But I don't understand the justification for changing the increment. What am I missing?

jintekc:
"Confirm Duplicate Update" only raises when the hashes **_differ_.**

> This step confirms that an update with a lower-than-expected targetVersionId is a true duplicate.
> 
> Create unsigned_update by removing the proof property from update. Hash unsigned_update with the [JSON Document Hashing](https://dcdpr.github.io/did-btcr2/algorithms.html#json-document-hashing) algorithm and compare it to update_hash_history[update.targetVersionId - 2]. Raise a [LATE_PUBLISHING](https://dcdpr.github.io/did-btcr2/errors.html#late_publishing) error if the hashes differ.

Issue #321 and this corresponding PR are primarily focused on the case of when hashes **_match_** because then the "Confirm Duplicate Update" sub-routine returns successfully and continues executing the outer loop, which would then execute the step "Increment `current_version_id`." This would be done erroneously since the update was not a true version bump, but rather a confirmed duplicate, likely resulting from an accidental broadcast of the same update more than once.

The whole point of the "Confirm Duplicate Update" step is to allow users the ability to make the mistake of broadcasting an identical update more than once without bricking their DID resolution on a "LATE PUBLISHING" error, since technically its not late publishing. The net side effect of how we wrote it, however, is that it will end up inflating `current_version_id` if we do not scope that step to the "Apply update" sub-section.

parasyte (approved):
That's what I'm missing. I've misread my own code! *An* increment happens in that location, but the result is stored in a separate variable and used for the condition on next line. This error was probably introduced by me when translating this code to the new spec prose. A misfortunate simplification.

Yes, the increment to `current_version_id` needs to be moved into "Apply `update`". That is a much closer match to the code that this description was based on.

----

I went through a whole comparison between the code and the old spec it was based on. That revealed a gap in the old spec that I believe the new spec with this PR adequately resolves. It doesn't matter for this PR, so I'll hide the analysis below. It might be useful to future archeologists.

<details><summary>Analysis of old spec...</summary>

The code I'm looking at was based on the pandoc version of the spec: https://github.com/dcdpr/did-btcr2/blob/75ae6934458922e70d24d8f3c22db423c7df017f/chapters/CRUD-Operations.md#traverse-bitcoin-blockchain-history.

Step 10 (you'll have to scroll down a page) is the process in question. Step 10.i is `update.targetVersionId <= current_version_id`, 10.ii is `update.targetVersionId == current_version_id + 1`, and 10.iii is **NOT** `update.targetVersionId > current_version_id + 1`, it is `... + 2`! `current_version_id` gets incremented within the `==` case.

The way this older spec (and the code that I'm comparing to) was written, the third condition requires the extra increment, and that leaves a gap where `update.targetVersionId > current_version_id + 1` is not actually checked. (In short, the old spec and the code are missing `else` clauses on the latter two conditions. The new spec fixes this ambiguity by stating "Only one of three possible conditions will occur:")

I do believe we need to check that gap (unlike the old spec). And I think that moving the increment into "Apply `update`" is the right thing to do (matching the old spec).
</details>

danpape (approved):

PR #338, opened 2026-07-28
Require a minimum confirmation count for beacon transactions

### LINKED ISSUE #37: Determine Target Blockheight - How many confirmations
The target blockheight defines the bitcoin block that the resolution algorithm goes up to.

It resolves the DID document for a btc1 DID at that targetted blockheight.

Currently, the resolution algorithm takes in a versionTime. This is a UNIX time value. We then determine what the targetBlockheight is at that time. That is fine.

But, if there is no versionTime, we set the targetBlockheight to the latest Bitcoin block with x number of confirmations.

Question: What is x? And is it variable? Currently I have been setting the min number of confirmations to 6, but it is not in the spec.

jintekc:
Adds the confirmation rule discussed in the issue: a beacon transaction MUST be included in a Bitcoin block and have at least resolutionOptions.minConf confirmations (default 6) to count towards resolution. minConf is floored at 1: mempool contents are node-local, unordered, and replaceable (RBF), so unconfirmed transactions never count regardless of options. A footnote explains that 6 is the broadly accepted settlement standard and that lowering minConf trades reorg exposure, which consumers can evaluate from the REQUIRED confirmations metadata. The Resolution Options data structure and example gain the new minConf property.

Closes #37

### DIFF
--- a/src/data-structures.md
+++ b/src/data-structures.md
@@ -283,6 +283,7 @@ Resolution options MAY contain the following properties:
 
 - `versionId`: OPTIONAL ASCII string representation of the specific version of a DID document to be resolved.
 - `versionTime`: OPTIONAL XML Datetime normalized to UTC without sub-second decimal precision. The DID document to be resolved is the most recent version of the DID document that was valid for the DID before the specified `versionTime`.
+- `minConf`: OPTIONAL positive integer (minimum `1`). The minimum number of Bitcoin block confirmations a [Beacon Signal] transaction requires to count towards resolution. Defaults to `6`.
 - `sidecar`: [Sidecar Data (data structure)].
 
 {% set hide_text = `` %}
--- a/src/data-structures.md
+++ b/src/data-structures.md
@@ -283,7 +283,7 @@ Resolution options MAY contain the following properties:
 
 - `versionId`: OPTIONAL ASCII string representation of the specific version of a DID document to be resolved.
 - `versionTime`: OPTIONAL XML Datetime normalized to UTC without sub-second decimal precision. The DID document to be resolved is the most recent version of the DID document that was valid for the DID before the specified `versionTime`.
-- `minConf`: OPTIONAL positive integer (minimum `1`). The minimum number of Bitcoin block confirmations a [Beacon Signal] transaction requires to count towards resolution. Defaults to `6`.
+- `minConf`: OPTIONAL positive integer (minimum `1`). The minimum number of Bitcoin block confirmations required on a [Beacon Signal] transaction during resolution. Defaults to `6`.
 - `sidecar`: [Sidecar Data (data structure)].
 
 {% set hide_text = `` %}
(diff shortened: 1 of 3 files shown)


danpape (dismissed):

src/operations/resolve.md
```diff
@@ -118,6 +118,10 @@ Scan the `service` entries in `current_document` ([DID Document (data structure)
 
 Implementations are RECOMMENDED to query an indexed Bitcoin blockchain Remote Procedure Call (RPC) service such as [electrs](https://github.com/romanz/electrs) or [Esplora](https://github.com/Blockstream/esplora). Implementations MAY instead traverse blocks from the genesis block. Cache [Beacon Addresses][Beacon Address] to avoid repeated transaction lookups.
 
+A transaction MUST be included in a Bitcoin block and have at least `resolutionOptions.minConf` confirmations (`6` when not provided) for it to count towards resolution. Unconfirmed mempool transactions MUST NOT be processed. [^2]
```
parasyte:
```suggestion
A transaction MUST be included in a Bitcoin block and have at least `resolutionOptions.minConf` confirmations (`6` when not provided). Unconfirmed mempool transactions MUST NOT be processed. [^2]
```

I don't think this epilogue phrase adds anything of value here.

src/data-structures.md
```diff
@@ -283,6 +283,7 @@ Resolution options MAY contain the following properties:
 
 - `versionId`: OPTIONAL ASCII string representation of the specific version of a DID document to be resolved.
 - `versionTime`: OPTIONAL XML Datetime normalized to UTC without sub-second decimal precision. The DID document to be resolved is the most recent version of the DID document that was valid for the DID before the specified `versionTime`.
+- `minConf`: OPTIONAL positive integer (minimum `1`). The minimum number of Bitcoin block confirmations a [Beacon Signal] transaction requires to count towards resolution. Defaults to `6`.
```
parasyte:
```suggestion
- `minConf`: OPTIONAL positive integer (minimum `1`). The minimum number of Bitcoin block confirmations required on a [Beacon Signal] transaction during resolution. Defaults to `6`.
```

I'm less confident about this rewording. I want to avoid the phrase "count towards" because it seems unusual for technical writing. I could see "considered for", "eligible for" and other variations. It feels like changing the sentence structure helps it read better without relying on "count".

My reasoning is that "count" is a mathematical term. The STE dictionary defines the word as "To add the number of objects or occurrences to get a total" and provides the approved example: "COUNT THE NUMBER OF TREADS ON THE TIRE." (On the other hand, the STE manual itself uses the phrase "counts as one word" 24 times. This suggests that I'm making a big deal over nothing.)

parasyte (dismissed):
Good change. I left some grammatical nitpicks.

src/data-structures.md
jintekc: lgtm. applying suggestion.

jintekc:
Grammatical nitpicks applied. Ready for re-review.

parasyte (approved):

danpape (approved):

Issue #339, opened 2026-08-03
Define the `INVALID_DID` error and fix its broken references


jintekc:
## Problem

`INVALID_DID` is raised in four places but defined nowhere:

* [algorithms.md:13](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/algorithms.md?plain=1#L13) and [algorithms.md:82](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/algorithms.md?plain=1#L82) link `INVALID_DID` as `(./errors.html)`, a raw link to the errors chapter, which has no matching section.
* [resolve.md:56](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/operations/resolve.md?plain=1#L56) and [resolve.md:68](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/operations/resolve.md?plain=1#L68) use the markdown reference ``[`INVALID_DID`]``, which has no definition in [errors-links.tera](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/includes/errors-links.tera#L1-L3) (it holds exactly three errors), so the built book renders literal brackets today.

## Solution

Add an `INVALID_DID` section to `errors.md`, register the link definition in `errors-links.tera`, and replace the two raw `(./errors.html)` links with the reference.

danpape:
Did we not define it because maybe we meant to refer to https://w3c.github.io/did-resolution/#errors ?

jintekc:
Yeah I guess we do have a small paragraph on the errors.html page

> The algorithms described in this specification can raise the specific errors listed below. Additional error types defined in Section 10 of DID Resolution v0.3 [[DID-RESOLUTION](https://dcdpr.github.io/did-btcr2/bibliography.html#DID-RESOLUTION)] may also be raised.

But if I were to land on that page as a implementer, I don't think I would continue digging. I would be confused by the lack of a reference to INVALID_DID. Maybe instead of `./errors.html` we link directly to https://w3c.github.io/did-resolution/#errors ?

jintekc:
Furthermore, we incorrectly reference Section 10. Errors are listed in Section 11 of v0.3 of DID-RESOLUTION.

jintekc:
## Updated Solution

[...]

Issue #329, opened 2026-07-24
Fix the CIDv1 byte recipe


jintekc:
## Problem

The CID recipe ([update-data-distribution.md:77-82](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/update-data-distribution.md?plain=1#L77-L82)) reads: "`0x01`, the code for CIDv1; `0x00`, the multicodec code for raw binary; `0x12`, the multihash code for SHA-256; and the SHA-256 of the file". Two byte errors:

1. `0x00` is the multicodec for `identity`; raw binary is `0x55` (which the spec's own Raw-Leaves mandate requires, [update-data-distribution.md:71-75](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/update-data-distribution.md?plain=1#L71-L75)).
2. A multihash is `<hash-code><digest-length><digest>` ([CIDv1 spec](https://specs.ipfs.tech/cid/#cidv1)); the digest-length byte `0x20` is missing.

The spec-literal CID `0x01 0x00 0x12 <32 bytes>` is malformed and no IPFS node recognizes it, so every CAS retrieval path fails as written ([resolve.md:68](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L68), [130-131](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L130-L131), [137](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L137)). Implementations noticing the error could also "fix" it differently, producing incompatible addresses for identical content.

(If you build CIDs with a multiformats library you won't see the length byte in your code; the multihash layer inserts it. That's why library-based implementations already emit the correct bytes and interoperate despite the spec text.)

## Solution

Correct the recipe to `0x01, 0x55, 0x12, 0x20, <digest>` and add one worked example (file bytes in, exact CID out) as a test vector. No migration cost: this matches what every shipped implementation already emits.

jintekc:
Closed by #340

Issue #328, opened 2026-07-24
Pin the DID Resolution draft and the resolver content types


jintekc:
Two small gaps in the HTTP-facing surface, best pinned before anyone ships a resolver endpoint.

### PROPOSALS

1. Pin the conformance target

* [conformance.md:14](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/conformance.md?plain=1#L14) requires conformance to "DID Resolution v0.3", a moving editor's draft: two implementers can target "v0.3" on different days and read different normative text.
* Proposal: cite an exact dated snapshot (date + URL) and update it deliberately.

2. State the resolver content types

* DID Resolution v0.3 returns a full resolution-result envelope as `application/did-resolution`; older tooling returned a bare document as `application/did+ld+json`. The spec mentions neither consistently: the prose example ([data-structures.md:312](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/data-structures.md?plain=1#L312)) says `application/ld+json` while its own example file ([did-resolution-metadata.json:2](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/example-data/did-resolution-metadata.json#L2)) says `application/did+ld+json`, and `application/did-resolution` appears nowhere.
* Proposal: state which content types a did:btcr2 resolver emits for the bare-document and full-envelope cases, and make the prose example agree with the example file.

danpape:
not to mention the DID resolution spec was just updated to "v1" (though still a draft): https://www.w3.org/TR/2026/WD-did-resolution-20260724/

jintekc:
Well shoot! I guess we should update to v1 in #340

jintekc:
Closed by #340

Issue #288, opened 2026-01-16
Need to add 'confirmations' to 'DID Document Metadata' in Data Structures


danpape:
Resolve -> Process mentions REQUIRED entries for 'DID Document Metadata' , but the data structure definition does not.

jintekc:
Closed by #340

PR #340, opened 2026-08-03
DID Document Metadata, ECMA-262 citation, CIDv1 bytes, `INVALID_DID`, DID Resolution pin

jintekc:
One commit per issue, none of the changes overlap. Bundled so it is one review pass instead of five.

- **#288**: defines `confirmations` in DID Document Metadata, makes `versionId` REQUIRED, and starts `block_confirmations` at `0` so a never-updated DID returns a defined value.
- **#167**: cites ECMA-262, 16th Edition at the three `encodeURIComponent()` sites (whole standard, per the issue thread).
- **#329**: corrects the CIDv1 bytes to `0x01, 0x55, 0x12, 0x20, <digest>` and adds a worked example as a test vector.
- **#339**: per the discussion there, DID Resolution already defines the resolution errors, so the four broken `INVALID_DID` references now link straight to its errors section instead of a new local definition.
- **#328**: cites the latest version of DID Resolution (`https://www.w3.org/TR/did-resolution/`) per the issue thread, saying "latest version" instead of naming one, and states the resolver content types: `application/did` and `application/did-resolution`, not `application/did+ld+json` as the issue suggested. DID Core v1.1 consolidated the media type to `application/did`.

### DIFF
--- a/src/operations/resolve.md
+++ b/src/operations/resolve.md
@@ -8,7 +8,7 @@
 
 Resolving a **did:btcr2** identifier iteratively builds a DID document by applying [BTCR2 Updates][BTCR2 Update] committed to the Bitcoin blockchain by [Authorized Beacon Signals][Authorized Beacon Signal] to an [Initial DID Document]. The [Initial DID Document] is either deterministically created from the DID or provided by [Sidecar Data].
 
-DID resolution is defined by DID Resolution v0.3 {{#cite DID-RESOLUTION}}.
+DID resolution is defined by DID Resolution v1 {{#cite DID-RESOLUTION}}.
 
 The resolve operation has the following function signature:
 
--- a/src/operations/resolve.md
+++ b/src/operations/resolve.md
@@ -28,7 +28,7 @@ Resolution maintains the following state while building the DID document:
 * `current_document`: the DID document being assembled.
 * `current_version_id`: the version number being processed (starts at `1`).
 * `update_hash_history`: a list of [BTCR2 Unsigned Update] hashes used to detect duplicates.
-* `block_confirmations`: confirmations for the Bitcoin block that contains the most recently applied unique update.
+* `block_confirmations`: confirmations for the Bitcoin block that contains the most recently applied unique update (starts at `0`).
 
 The resolver:
 
--- a/src/update-data-distribution.md
+++ b/src/update-data-distribution.md
@@ -1,3 +1,4 @@
+{% import "includes/ui.tera" as ui %}
 {% import "includes/links.tera" as links %}
 
 {{ links::include() }}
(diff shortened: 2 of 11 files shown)


danpape (dismissed):
I think everything looks great, especially the work on the CIDv1 inconsistency. I think the only thing I might question is the wishy-washy reference to the "latest" version of the DID Resolution spec. I guess that's fine for now since that spec is still evolving, but definitely before we stamp this spec as finalized, we should make sure that our dependencies point to finalized versions of other docs.

src/update-data-distribution.md
```diff
@@ -77,13 +78,37 @@ the [CID].
 The IPFS CIDv1 is a binary identifier constructed from the file hash as:
 
 * `0x01`, the code for CIDv1;
-* `0x00`, the [multicodec](https://github.com/multiformats/multicodec) code for raw binary;
-* `0x12`, the [multihash](https://github.com/multiformats/multihash) code for SHA-256; and
+* `0x55`, the [multicodec](https://github.com/multiformats/multicodec) code for raw binary;
+* `0x12`, the [multihash](https://github.com/multiformats/multihash) code for SHA-256;
+* `0x20`, the length of the SHA-256 digest (32 bytes); and
```
parasyte, line 83: How did that happen? These byte definitions were originally added way back in #84.

The changes seem to check out. It was not easy to find the expected byte sequence, but this seems to be it: https://specs.ipfs.tech/cid/#decoding-a-binary-cid The `<digest-length>` is definitely required.

I believe `0x55` is the correct multicodec byte for raw binary, assuming that the `ipld` tag specifies the bytes that are valid to use here. https://github.com/multiformats/multicodec#tag-categories seems to imply as much. (IMHO, the choice of a global namespace for registering unrelated scopes distinguished by tag categories is a mistake, but it's not my specification.)

src/update-data-distribution.md
```diff
@@ -77,13 +78,37 @@ the [CID].
...
+'
+Example input:
+
+* file content: the 3-byte ASCII string `"abc"` (`0x61 0x62 0x63`)
+
+Example output:
+
+* SHA-256 of the file: `ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad`
+* CIDv1 bytes: `01551220ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad`
+* stringified CIDv1 (multibase `base32`, prefix `b`): `"bafkreif2pall7dybz7vecqka3zo24irdwabwdi4wc55jznaq75q7eaavvu"`
```
parasyte, line 101: Example looks good according to CID Inspector: https://cid.ipfs.tech/#bafkreif2pall7dybz7vecqka3zo24irdwabwdi4wc55jznaq75q7eaavvu

src/operations/resolve.md
```diff
@@ -8,7 +8,7 @@
 
 Resolving a **did:btcr2** identifier iteratively builds a DID document by applying [BTCR2 Updates][BTCR2 Update] committed to the Bitcoin blockchain by [Authorized Beacon Signals][Authorized Beacon Signal] to an [Initial DID Document]. The [Initial DID Document] is either deterministically created from the DID or provided by [Sidecar Data].
 
-DID resolution is defined by DID Resolution v0.3 {{#cite DID-RESOLUTION}}.
+DID resolution is defined by the latest version of DID Resolution {{#cite DID-RESOLUTION}}.
```
parasyte: I dislike this change. Are you able to guarantee that every future version of the DID Resolution specification will always be compatible with what this specification requires?

parasyte (changes requested):
I strongly recommend defining precisely which version of the DID Resolution specification is being referenced.

src/operations/resolve.md
jintekc: sounds good. fixed.

jintekc:
DID Resolution spec version precisely pinned. @parasyte back to you for final review and approval.

danpape (approved):

parasyte (approved):

PR #345, opened 2026-08-12
Update semantics: single RFC 6902 patch document, patch-failure error, `targetVersionId` referent

### LINKED ISSUE #220: Error needed for bad JSON Patches?
We noticed the spec never says anything about failures due to a bad Patch. I suppose it could just roll up to a INVALID_DID_UPDATE, but maybe we want something more specific?

### LINKED ISSUE #322: Define `patch` as a single RFC 6902 patch document
The spec describes the update patch input two incompatible ways:

- `operations/update.md` (function signature and inputs): `jsonPatches`, "an array of JSON patch documents", later "serialized to a JSON string".
- `data-structures.md` (BTCR2 Unsigned Update): `patch` is "a JSON Patch object", i.e. a single RFC 6902 patch document.
- The template (`btcr2-unsigned-update-template.hbs`) splices `"patch": {{array-of-patches}}` and every example file shows `patch` as a flat array of operation objects.

An RFC 6902 patch document is itself an array of operations, so "array of JSON patch documents" implies an array of arrays on the wire, which no example shows and no implementation shipped.

Recommendation (proposed for discussion):

1. State the wire shape once, in `data-structures.md`: `patch` is a single RFC 6902 JSON Patch document (a flat array of operation objects).
2. In `operations/update.md`, either rename the input to `jsonPatch` (single document) or keep accepting several documents but state that the function concatenates their operations into the single wire `patch`.
3. Fix "serialized to a JSON string": the patch is embedded as JSON, not as a string value.

All three known implementations already use the single-flat-patch reading, so this is an editorial/normative-text fix with no compatibility impact.

### LINKED ISSUE #334: Clarify how `targetVersionId` is validated
## Problem

`update()` takes `targetVersionId` as an explicit input ([update.md:16-24](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/update.md?plain=1#L16-L24), [update.md:30](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/update.md?plain=1#L30)), but [data-structures.md:105-106](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/data-structures.md?plain=1#L105-L106) says the value MUST be one more than the source document's version. That MUST is unenforceable at construction: a DID document carries no version (version is resolution metadata), so neither the function nor a library can check it, and implementers are left guessing whether to validate, derive, or pass through. Getting it wrong is destructive: announcing an update whose `targetVersionId` exceeds current + 1 sends resolution into the gap branch and a fatal `LATE_PUBLISHING` ([resolve.md:160-169](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L160-L169)), made permanent by the on-chain signal.

## Proposal

State that `targetVersionId` is caller-supplied and enforced only at resolution time, and reword the data-structures MUST accordingly (the constraint is on what resolves, not on what constructs). Add a warning to `update.md`: callers SHOULD derive `targetVersionId` from a fresh resolve, because announcing a wrong value permanently invalidates the DID.

Welcoming feedback and discussion for alternative ideas/solutions.

jintekc:
One commit per issue. All three land in the same few lines of the update path, so bundling them is one review pass instead of three rebases.

- **#322**: the wire shape is now stated once, in Data Structures: `patch` is a single RFC 6902 patch document, one flat array of operation objects. `update()`'s input is renamed `jsonPatches` to `jsonPatch` to match, the patch is embedded in the template as JSON rather than "serialized to a JSON string", and the terminology entry drops the stray "object". The `{{array-of-patches}}` template variable name is left alone; renaming it would be cosmetic churn. All shipped implementations already read the patch this way, so nothing breaks.
- **#220**: a malformed patch or a failed operation (including a failed `test`) now raises `INVALID_DID_UPDATE`, per the thread. Defined at both places a patch is applied: `update()` construction and the resolve path's Apply `update` step, since an announced update is untrusted input there.
- **#334**: the old MUST said `targetVersionId` is "one more than the `versionId` of the DID document being updated", but DID documents carry no `versionId`; version is reported in DID Document Metadata. The rule now names its referent and states that it is enforced at resolution time. Deriving the value is part of the update process, so it gets its own normative paragraph in Construct BTCR2 Unsigned Update: take it from a fresh resolution rather than a locally maintained count, because announcing a wrong value in either direction can permanently prevent the DID from resolving. `update()`'s input list keeps only the definition.

Fixes #322
Fixes #220
Fixes #334

### DIFF
--- a/src/data-structures.md
+++ b/src/data-structures.md
@@ -103,8 +103,12 @@ SHA-256 hashes {{#cite SHA256}} (`targetHash` and `sourceHash`) MUST be produced
   operation objects, that defines a set of transformations to be applied to a DID document. The
   result of applying the patch MUST be a conformant DID document according to the DID core v1.1
   specification {{#cite DID-CORE}}.
-- `targetVersionId`: The `versionId` of the DID document after the patch has been applied. The
-  targetVersionId MUST be one more than the `versionId` of the DID document being updated.
+- `targetVersionId`: The version of the DID document after the patch has been applied, i.e., the
+  `versionId` that will be returned in the [DID document metadata (data structure)] once this
+  update is applied. The `targetVersionId` MUST be one more than the integer form of the
+  `versionId` returned in the [DID document metadata (data structure)] before this update is
+  applied. `versionId` is never a property of the DID document itself, so this requirement
+  cannot be checked when the update is constructed; it is enforced at resolution time.
 - `sourceHash`: SHA-256 hash of the DID document that the patch MUST be applied to. The hash MUST be produced by the [JSON Document Hashing] algorithm.
 - `targetHash`: SHA-256 hash of the DID document that results from applying the patch to the source document. The hash MUST be produced by the [JSON Document Hashing] algorithm.
 
--- a/src/data-structures.md
+++ b/src/data-structures.md
@@ -104,10 +104,10 @@ SHA-256 hashes {{#cite SHA256}} (`targetHash` and `sourceHash`) MUST be produced
   result of applying the patch MUST be a conformant DID document according to the DID core v1.1
   specification {{#cite DID-CORE}}.
 - `targetVersionId`: The version of the DID document that results from applying the patch, i.e., the
-`versionId` that will be returned in the [DID document metadata (data structure)] for the updated DID 
-document. The `targetVersionId` MUST be one more than the integer form of the `versionId` of the DID 
-document being updated. `versionId` is never a property of the DID document itself, so this requirement 
-cannot be checked when the update is constructed; it is enforced at resolution time.
+  `versionId` that will be returned in the [DID document metadata (data structure)] for the updated DID 
+  document. The `targetVersionId` MUST be one more than the integer form of the `versionId` of the DID 
+  document being updated. `versionId` is never a property of the DID document itself, so this requirement 
+  cannot be checked when the update is constructed; it is enforced at resolution time.
 - `sourceHash`: SHA-256 hash of the DID document that the patch MUST be applied to. The hash MUST be produced by the [JSON Document Hashing] algorithm.
 - `targetHash`: SHA-256 hash of the DID document that results from applying the patch to the source document. The hash MUST be produced by the [JSON Document Hashing] algorithm.
 
--- a/src/data-structures.md
+++ b/src/data-structures.md
@@ -99,9 +99,10 @@ SHA-256 hashes {{#cite SHA256}} (`targetHash` and `sourceHash`) MUST be produced
   - `"https://w3id.org/security/data-integrity/v2"`
   - `"https://w3id.org/json-ld-patch/v1"`
   - `"https://btcr2.dev/context/v1"`
-- `patch`: A JSON Patch {{#cite RFC6902}} object that defines a set of transformations to be
-  applied to a DID document. The result of applying the patch MUST be a conformant DID document
-  according to the DID core v1.1 specification {{#cite DID-CORE}}.
+- `patch`: A single JSON Patch {{#cite RFC6902}} document, i.e., one flat array of JSON Patch
+  operation objects, that defines a set of transformations to be applied to a DID document. The
+  result of applying the patch MUST be a conformant DID document according to the DID core v1.1
+  specification {{#cite DID-CORE}}.
 - `targetVersionId`: The `versionId` of the DID document after the patch has been applied. The
   targetVersionId MUST be one more than the `versionId` of the DID document being updated.
 - `sourceHash`: SHA-256 hash of the DID document that the patch MUST be applied to. The hash MUST be produced by the [JSON Document Hashing] algorithm.
--- a/src/data-structures.md
+++ b/src/data-structures.md
@@ -103,12 +103,11 @@ SHA-256 hashes {{#cite SHA256}} (`targetHash` and `sourceHash`) MUST be produced
   operation objects, that defines a set of transformations to be applied to a DID document. The
   result of applying the patch MUST be a conformant DID document according to the DID core v1.1
   specification {{#cite DID-CORE}}.
-- `targetVersionId`: The version of the DID document after the patch has been applied, i.e., the
-  `versionId` that will be returned in the [DID document metadata (data structure)] once this
-  update is applied. The `targetVersionId` MUST be one more than the integer form of the
-  `versionId` returned in the [DID document metadata (data structure)] before this update is
-  applied. `versionId` is never a property of the DID document itself, so this requirement
-  cannot be checked when the update is constructed; it is enforced at resolution time.
+- `targetVersionId`: The version of the DID document that results from applying the patch, i.e., the
+`versionId` that will be returned in the [DID document metadata (data structure)] for the updated DID 
+document. The `targetVersionId` MUST be one more than the integer form of the `versionId` of the DID 
+document being updated. `versionId` is never a property of the DID document itself, so this requirement 
+cannot be checked when the update is constructed; it is enforced at resolution time.
 - `sourceHash`: SHA-256 hash of the DID document that the patch MUST be applied to. The hash MUST be produced by the [JSON Document Hashing] algorithm.
 - `targetHash`: SHA-256 hash of the DID document that results from applying the patch to the source document. The hash MUST be produced by the [JSON Document Hashing] algorithm.
 
--- a/src/operations/update.md
+++ b/src/operations/update.md
@@ -15,7 +15,7 @@ The update operation has the following function signature:
 ```rust
 fn update(
   didSourceDocument,
-  jsonPatches,
+  jsonPatch,
   targetVersionId,
   verificationMethodId,
   privateKey,
@@ -26,7 +26,7 @@ fn update(
 Input arguments:
 
 - `didSourceDocument`: The source DID document.
-- `jsonPatches`: An array of JSON patch documents {{#cite RFC6902}} with the changes to be made to the source DID document.
+- `jsonPatch`: A single JSON Patch document {{#cite RFC6902}} with the changes to be made to the source DID document. Its wire shape is defined by the `patch` property of the [BTCR2 Unsigned Update (data structure)].
 - `targetVersionId`: The `versionId` of the target DID document that the new [BTCR2 Signed Update] will yield.
 - `verificationMethodId`: The `verificationMethod` ID used for signing the [BTCR2 Update].
 - `privateKey`: Private key associated with the `verificationMethodId`.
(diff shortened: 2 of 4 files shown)


src/operations/update.md
```diff
@@ -26,8 +26,8 @@ fn update(
 Input arguments:
 
 - `didSourceDocument`: The source DID document.
-- `jsonPatches`: An array of JSON patch documents {{#cite RFC6902}} with the changes to be made to the source DID document.
-- `targetVersionId`: The `versionId` of the target DID document that the new [BTCR2 Signed Update] will yield.
+- `jsonPatch`: A single JSON Patch document {{#cite RFC6902}} with the changes to be made to the source DID document. Its wire shape is defined by the `patch` property of the [BTCR2 Unsigned Update (data structure)].
+- `targetVersionId`: The `versionId` that resolutions of the DID will report once the new [BTCR2 Signed Update] is applied. Callers SHOULD derive this value from a fresh resolution of the DID, as one more than the integer form of the `versionId` reported in the [DID document metadata (data structure)], rather than keeping their own count. The construction process performs no check on this value; it is enforced at resolution time, and announcing a [BTCR2 Signed Update] whose `targetVersionId` is wrong in either direction can permanently prevent the DID from resolving.
```
danpape: I'm not sure I like the use of 'resolution' as a noun. How about: 


```suggestion
- `targetVersionId`: The `versionId` that will be returned in the [DID document metadata (data structure)] once the new [BTCR2 Signed Update] is applied. Callers SHOULD derive this value from a fresh resolution of the DID, as one more than the integer form of the `versionId` reported in the [DID document metadata (data structure)], rather than keeping their own count. The construction process performs no check on this value; it is enforced at resolution time, and announcing a [BTCR2 Signed Update] whose `targetVersionId` is wrong in either direction can permanently prevent the DID from resolving.
```

danpape (changes requested):

src/operations/update.md
```diff
@@ -26,8 +26,8 @@ fn update(
 Input arguments:
 
 - `didSourceDocument`: The source DID document.
-- `jsonPatches`: An array of JSON patch documents {{#cite RFC6902}} with the changes to be made to the source DID document.
-- `targetVersionId`: The `versionId` of the target DID document that the new [BTCR2 Signed Update] will yield.
+- `jsonPatch`: A single JSON Patch document {{#cite RFC6902}} with the changes to be made to the source DID document. Its wire shape is defined by the `patch` property of the [BTCR2 Unsigned Update (data structure)].
+- `targetVersionId`: The `versionId` that will be returned in the [DID document metadata (data structure)] once the new [BTCR2 Signed Update] is applied. Callers SHOULD derive this value from a fresh resolution of the DID, as one more than the integer form of the `versionId` reported in the [DID document metadata (data structure)], rather than keeping their own count. The construction process performs no check on this value; it is enforced at resolution time, and announcing a [BTCR2 Signed Update] whose `targetVersionId` is wrong in either direction can permanently prevent the DID from resolving.
```
parasyte: I would probably put the rule for deriving `targetVersionId` in its own normative paragraph. This specific section is defining the inputs, not necessarily the update semantics.

src/operations/update.md
```diff
@@ -48,11 +48,11 @@ Constructing a [BTCR2 Signed Update] is a two-step process. First, a [BTCR2 Unsi
 
 This process constructs a [BTCR2 Unsigned Update (data structure)].
 
-Apply all JSON patches in `jsonPatches` to `didSourceDocument` to create `didTargetDocument`. `didTargetDocument` MUST be conformant to DID Core v1.1 {{#cite DID-CORE}}. An [`INVALID_DID_UPDATE`] error MUST be raised if `didTargetDocument.id` is not equal to `didSourceDocument.id`.
+Apply `jsonPatch` to `didSourceDocument` to create `didTargetDocument`. An [`INVALID_DID_UPDATE`] error MUST be raised if `jsonPatch` is malformed or fails to apply. (JSON Patch {{#cite RFC6902}} operations are evaluated in order; the first operation that fails, including a failed `test` operation, fails the whole patch.) `didTargetDocument` MUST be conformant to DID Core v1.1 {{#cite DID-CORE}}. An [`INVALID_DID_UPDATE`] error MUST be raised if `didTargetDocument.id` is not equal to `didSourceDocument.id`.
```
parasyte: Should this note about patch order/testing be omitted here, or added to resolve.md:190? For consistency between the two places where JSON patches are applied.

Alternatively, a short section in the Algorithms chapter can cover JSON patches, if appropriate.

src/operations/update.md
```diff
@@ -48,11 +48,11 @@ Constructing a [BTCR2 Signed Update] is a two-step process. First, a [BTCR2 Unsi
...
 
 Fill the [BTCR2 Unsigned Update (data structure)] template below with the required template variables.
 
-* `array-of-patches`: `jsonPatches` serialized to a JSON string.
+* `array-of-patches`: `jsonPatch` embedded as JSON. The patch becomes the array value of the `patch` property, not a JSON string value.
```
parasyte: Remove the "not a JSON string value". This "retain the negative" behavior is an annoying LLM-ism when they make corrections.

It's basically saying, "don't do it the wrong way like it was written before in a revision of the document you didn't even read and were completely unaware of until I brought it up for no reason at all".

src/data-structures.md
```diff
@@ -99,11 +99,17 @@ SHA-256 hashes {{#cite SHA256}} (`targetHash` and `sourceHash`) MUST be produced
...
+  operation objects, that defines a set of transformations to be applied to a DID document. The
+  result of applying the patch MUST be a conformant DID document according to the DID core v1.1
+  specification {{#cite DID-CORE}}.
+- `targetVersionId`: The version of the DID document after the patch has been applied, i.e., the
+  `versionId` that resolutions of the DID will report in the
+  [DID document metadata (data structure)] once this update is applied. The `targetVersionId`
+  MUST be one more than the integer form of the `versionId` reported when resolving the DID
+  before this update is applied. `versionId` is never a property of the DID document itself, so
+  this requirement cannot be checked when the update is constructed; it is enforced at resolution
+  time.
```
parasyte: I see, the other change to the Update chapter is now repeating this semantic part from the Data Structures chapter. Even in this one PR, both places say different things about the semantic interpretation of the `targetVersionId`.

The Update chapter says "SHOULD derive this value from a fresh resolution ..." and this one says "MUST be one more than the integer form of the `versionId` reported when resolving ..." These are different conditions with differing normative strengths.

For this reason alone, I prefer to keep redundancies to an absolute minimum. The more we can use links to cross-reference semantics, the better.

parasyte (changes requested):

src/operations/update.md
jintekc: yeah fair enough. I think it makes sense to move it to the Construct Unsigned Update section since its meant to be part of the update process.

src/operations/update.md
jintekc: I think it has value being there, so I agree, adding it to resolve for consistency makes sense.

src/data-structures.md
jintekc: Good catch. Removing the redundant and conflicting paragraph in update.md in favor of a more verbose version in data-structures.md.

src/operations/update.md
```diff
@@ -48,11 +48,13 @@ Constructing a [BTCR2 Signed Update] is a two-step process. First, a [BTCR2 Unsi
 
 This process constructs a [BTCR2 Unsigned Update (data structure)].
 
-Apply all JSON patches in `jsonPatches` to `didSourceDocument` to create `didTargetDocument`. `didTargetDocument` MUST be conformant to DID Core v1.1 {{#cite DID-CORE}}. An [`INVALID_DID_UPDATE`] error MUST be raised if `didTargetDocument.id` is not equal to `didSourceDocument.id`.
+Apply `jsonPatch` to `didSourceDocument` to create `didTargetDocument`. An [`INVALID_DID_UPDATE`] error MUST be raised if `jsonPatch` is malformed or fails to apply. JSON Patch {{#cite RFC6902}} operations are evaluated in order; the first operation that fails, including a failed `test` operation, fails the whole patch. `didTargetDocument` MUST be conformant to DID Core v1.1 {{#cite DID-CORE}}. An [`INVALID_DID_UPDATE`] error MUST be raised if `didTargetDocument.id` is not equal to `didSourceDocument.id`.
+
+`targetVersionId` MUST be taken from a fresh resolution of the DID rather than from a locally maintained count. Announcing a [BTCR2 Signed Update] whose `targetVersionId` is wrong in either direction can permanently prevent the DID from resolving.
```
parasyte: Should this say something specifically about `versionId` in the DID document metadata? It's implied, but I don't feel that's strong enough. (That wording still exists in the Data Structures chapter for the unsigned update.)

jintekc: Yeah that's a fair point. This should be more explicit. Maybe something like 

> `targetVersionId` MUST be derived from the `versionId` returned in the [DID document metadata (data structure)] by a fresh resolution of the DID ...

I'll push up an edit

src/data-structures.md
```diff
@@ -99,11 +99,16 @@ SHA-256 hashes {{#cite SHA256}} (`targetHash` and `sourceHash`) MUST be produced
...
+- `patch`: A single JSON Patch {{#cite RFC6902}} document, i.e., one flat array of JSON Patch
+  operation objects, that defines a set of transformations to be applied to a DID document. The
+  result of applying the patch MUST be a conformant DID document according to the DID core v1.1
+  specification {{#cite DID-CORE}}.
+- `targetVersionId`: The version of the DID document after the patch has been applied, i.e., the
+  `versionId` that will be returned in the [DID document metadata (data structure)] once this
+  update is applied. The `targetVersionId` MUST be one more than the integer form of the
+  `versionId` returned in the [DID document metadata (data structure)] before this update is
+  applied. `versionId` is never a property of the DID document itself, so this requirement
+  cannot be checked when the update is constructed; it is enforced at resolution time.
```
danpape:
```suggestion
- `targetVersionId`: The version of the DID document that results from applying the patch, i.e., the
`versionId` that will be returned in the [DID document metadata (data structure)] for the updated DID 
document. The `targetVersionId` MUST be one more than the integer form of the `versionId` of the DID 
document being updated. `versionId` is never a property of the DID document itself, so this requirement 
cannot be checked when the update is constructed; it is enforced at resolution time.
```

just trying to get rid of what seemed like too many "is applied" statements.

danpape (changes requested):
Looks good, but I have a suggestion for one part.

src/data-structures.md
parasyte: I noticed this patch removed the two-space indentation, which is inconsistent with the other list items.

danpape: oops

src/data-structures.md
```diff
@@ -99,11 +99,15 @@ SHA-256 hashes {{#cite SHA256}} (`targetHash` and `sourceHash`) MUST be produced
...
-  targetVersionId MUST be one more than the `versionId` of the DID document being updated.
+- `patch`: A single JSON Patch {{#cite RFC6902}} document, i.e., one flat array of JSON Patch
+  operation objects, that defines a set of transformations to be applied to a DID document. The
+  result of applying the patch MUST be a conformant DID document according to the DID core v1.1
+  specification {{#cite DID-CORE}}.
+- `targetVersionId`: The version of the DID document that results from applying the patch, i.e., the
+`versionId` that will be returned in the [DID document metadata (data structure)] for the updated DID 
+document. The `targetVersionId` MUST be one more than the integer form of the `versionId` of the DID 
+document being updated. `versionId` is never a property of the DID document itself, so this requirement 
+cannot be checked when the update is constructed; it is enforced at resolution time.
```
danpape:
```suggestion
- `targetVersionId`: The version of the DID document that results from applying the patch, i.e., the
  `versionId` that will be returned in the [DID document metadata (data structure)] for the updated DID 
  document. The `targetVersionId` MUST be one more than the integer form of the `versionId` of the DID 
  document being updated. `versionId` is never a property of the DID document itself, so this requirement 
  cannot be checked when the update is constructed; it is enforced at resolution time.
```

danpape (changes requested):

parasyte (approved):

danpape (approved):

Issue #331, opened 2026-07-24
Define how custom-network agreement works


jintekc:
Table 1 allocates `12`..`15` to "Custom networks" ([algorithms.md:33-34](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/algorithms.md?plain=1#L33-L34)). The spec defines no method to identify or to agree on a custom network. The nibble carries no chain identity. Two resolvers that are both "configured for" value `12` can bind the value to different chains. Then the two resolvers resolve the same DID against different blockchains. Neither resolver raises an error.

Custom networks occur in practice. `mutinynet` was a custom network before it got value `5` (#70). Three slots guarantee collisions between unrelated deployments.

jintekc:
PR #346 added a mutual-agreement paragraph ([algorithms.md:38-41](https://github.com/dcdpr/did-btcr2/blob/be67bce/src/algorithms.md?plain=1#L38-L41)) and a decode rule for custom values ([algorithms.md:103](https://github.com/dcdpr/did-btcr2/blob/be67bce/src/algorithms.md?plain=1#L103)). I think that is sufficient to cover most of this issue. The remaining gap is a chain identity for a custom network; however, in this case, the custom networks in question are simply test networks making this issue low impact.

Issue #301, opened 2026-03-11
Encoding Unclear for sourceHash and targetHash in BTCR2 Update


jintekc:
Thanks to @wip-abramson for pointing out that the current spec does not specify what the encoding should be for the `sourceHash` and `targetHash` values in a BTCR2. See [BTCR2 Unsigned Update](https://dcdpr.github.io/did-btcr2/data-structures.html#btcr2-unsigned-update) for details.

I propose using [base58btc](https://github.com/multiformats/multibase?tab=readme-ov-file#multibase-table). should include a leading "z"  The leading "z" is a multibase encoding convention and not native to bitcoin base58 encoding. However, we do specify that the `proofValue` field of a [Data Integrity Proof](https://dcdpr.github.io/did-btcr2/data-structures.html#data-integrity-proof-example-panel-show) be encoded "as a Multibase "base-58-btc" [[CID](https://dcdpr.github.io/did-btcr2/bibliography.html#CID)] encoded string."

We need to align all three to have the same encoding: either base58btc or something else.

I think it would greatly simplify things if we specify the encoding for all three (`sourceHash`, `targetHash`, `proofValue`) to be "base-58-btc" - a base58 string using multibase encoding.

danpape:
Given that, should we change proofValue to also be base64, or should we convert all three to be base58btc?

parasyte:
Both are already specified. Is this about proposing a change to how these are serialized?

My personal preference would be base64 everywhere (since we can’t have binary, base64 is the next best thing). I think base58 for the proofValue is a compromise because base58 was already chosen “for us” by the Data Integrity BIP340 Cryptosuites v0.1 speciation.

Issue #219, opened 2025-09-04
Tracking Major Spec Changes


wip-abramson:
This issue tracks major changes to the did:btc1 specification that will require implementation updates across existing implementations in order to align with the spec.

These changes are:

- Use base64 encoding instead of base58 for source and target hash of BTC1 Update (https://dcdpr.github.io/did-btc1/#construct-btc1-update)
- Update the context URL used in the DID document to https://btc1.dev/context/v1. 

There are also other major changes under discussion:

1. Migrate from JCS to RDFC for the cryptosuite
2. Defining BTC1 Beacons in DID document #170 
3. Sidecar Data restructure #178

PR #122, opened 2025-08-07
Various edits throughout the specification

jintekc:
Addresses the following issues
* #35 
* #58
* #98 
* #104
* #106 
* #114  
* #121
* #128
* #130
* #163

PR #346, opened 2026-08-12
Canonical DID string: lowercase and byte-conversion rules, network-value decode behavior

### LINKED ISSUE #330: Make the DID string canonical
## Problem

One identifier can have several distinct, checksum-valid DID strings:

1. Padding bits. The encode step says only "Encode the unencoded data bytes with Bech32m" ([algorithms.md:51](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/algorithms.md?plain=1#L51)). BIP350 defines the checksum, not the 8-to-5 repacking: the payload is 33 or 34 bytes, leaving 1-3 trailing pad bits, and nothing requires them to be zero or requires decoders to reject non-zero padding (BIP173's rule, never cited here).
2. Case. Bech32m is case-insensitive; DID syntax is case-sensitive. Only the first ten characters are pinned ([algorithms.md:86](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/algorithms.md?plain=1#L86)), so `did:btcr2:K1Q…` and `did:btcr2:k1q…` are both valid.

## Demonstration

Same pubkey (the secp256k1 generator point), `mutinynet`, pad bits `000` vs `101`:

```
did:btcr2:k1q5p8n0nx0muaewav2ksx99wwsu9swq5mlndjmn3gm9vl9q2mzmup0xqr4e30t   (pad bits 000)
did:btcr2:k1q5p8n0nx0muaewav2ksx99wwsu9swq5mlndjmn3gm9vl9q2mzmup0x93f0zar   (pad bits 101)
```

Both pass the Bech32m checksum and decode to the same 34 payload bytes. A BIP173 decoder rejects the second ("non-zero padding"); a decoder implementing only what the spec cites accepts both, so the `current_document.id == did` check ([resolve.md:187](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L187)) diverges by resolver.

## Solution

Require BIP173 `convertbits` with zero padding, require decoders to reject non-zero padding bits, and require the `method-specific-id` to be lowercase. One identifier, one string.

### LINKED ISSUE #332: Define decode behavior for reserved, custom, and unassigned network values
## Problem

Decode requires the network nibble to "be one of the values in Table 1" ([algorithms.md:91](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/algorithms.md?plain=1#L91)), but `Reserved` `6`..`11` and `Custom networks` `12`..`14` are rows in Table 1 ([algorithms.md:33-34](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/algorithms.md?plain=1#L33-L34)), so they satisfy the MUST while mapping to no `network_name`; value `15` is absent from the table entirely. Resolvers can disagree on whether such a DID is valid-but-unnamed, an error, or undefined behavior.

## Solution

State it explicitly: reserved values rejected until assigned; custom values rejected unless the resolver is configured for that network; `15` rejected. (Or narrow the MUST to the named values and define the error.) Also tighten the nibble-packing prose so encode and decode read as exact inverses.

Also, see the separate but related issue #331 to discuss how custom-network configuration works.

jintekc:
Two commits, one per issue, both in the identifier encode/decode chapter.

- **#330**: BIP350 defines the Bech32m checksum but not the 8-to-5 bit repacking, so the trailing pad bits were unconstrained, and Bech32m is case-insensitive while DID syntax is case-sensitive. One identifier had several distinct checksum-valid DID strings; the issue demonstrates two that decode to the same 34 payload bytes. The `method-specific-id` is now required to be lowercase, and a footnote shared by the encode and decode algorithms states the zero-padding rule directly, citing BIP350 for Bech32m itself. One identifier, one string. Adds the BIP173 bibtex entry. Also regenerates the encoding example, whose input key was not a valid SEC key and whose output carried a `did:btc1:` prefix: the new vector uses the secp256k1 generator point on `bitcoin` and was cross-checked by two independent implementations. The decoding example was verified already canonical and is unchanged.
- **#332**: the `Reserved` and `Custom networks` rows satisfied the old decode rule ("MUST be one of the values in Table 1") while mapping to no `network_name`. Decode now handles each row explicitly: named values map, reserved values are rejected until assigned (and must not be encoded), and custom networks, realigned to the old spec's `C`..`F` range (`12`..`15`), keep the old spec's framing: interpretation by mutual agreement between the parties issuing and resolving the identifier, no interoperability guarantee, and implementations SHOULD reject custom values they do not support. Also states that decoding inverts encoding exactly, which #330's padding and case rules make true.

Fixes #330
Fixes #332

### DIFF
--- a/src/algorithms.md
+++ b/src/algorithms.md
@@ -31,7 +31,14 @@ The `network_name` value declares which Bitcoin network anchors the identifier. 
 | `testnet4`       | `4`             |
 | `mutinynet`      | `5`             |
 | Reserved         | `6`..`11`       |
-| Custom networks  | `12`..`14`      |
+| Custom networks  | `12`..`15`      |
+
+`Reserved` network values MUST NOT be encoded until this specification assigns them.
+
+A `Custom networks` value allows any user to stand up a custom network and create **did:btcr2** identifiers on it. Anyone encountering such an identifier has to know the details of the network (e.g., the challenge and seed nodes for a custom signet) to use it. This means:
+
+* The interpretation of a custom value is by mutual agreement between the parties issuing **did:btcr2** identifiers and those resolving them.
+* Other users may use the same value, and their **did:btcr2** identifiers will **not** be interoperable in each other's networks.
 
 Introduce `btcr2_version` as `version_number - 1` (i.e., `0`). Combine `btcr2_version` in the low [^1] nibble and `network_value` in the high [^1] nibble into a single byte value.
 
@@ -90,12 +97,17 @@ A **did:btcr2** identifier MUST be processed according to the DID Resolution Alg
 The `method-specific-id` MUST be lowercase. Decode it as a Bech32m encoded string [^3] to retrieve the unencoded data bytes and the `hrp`. Parse the unencoded data bytes according to [Table 2: Unencoded Data Bytes](#unencoded-data-bytes) to retrieve the `btcr2_version`, `network_value` and `genesis_bytes`.
 
 * `btcr2_version` MUST be `0`. Introduce `version_number` as `btcr2_version + 1`. I.e., `version_number` MUST be returned to the caller as a type that can be represented as the value `1`.
-* `network_value` MUST be one of the values in [Table 1: Network Values](#network-values). `network_name` SHOULD be returned to the caller including additional type information using symbolic names that can be represented as integer values.
+* `network_value` MUST be handled according to its row in [Table 1: Network Values](#network-values):
+  * A named network value (`0`..`5`) maps to its `network_name`. `network_name` SHOULD be returned to the caller including additional type information using symbolic names that can be represented as integer values.
+  * A `Reserved` value (`6`..`11`) MUST be rejected until this specification assigns it a `network_name`.
+  * A `Custom networks` value (`12`..`15`) SHOULD be rejected when the implementation does not support a custom network for that value. What it means to support a custom network is implementation-defined.
 * The `hrp` MUST be either `"k"` or `"x"`.
 
 If the `hrp` is `"k"` (key-based **btcr2:did** identifier), `key_or_hash` MUST be `genesis_bytes` interpreted as a 33-byte SEC encoded secp256k1 public key.
 
 If the `hrp` is `"x"` ([Genesis Document]-based **btcr2:did** identifier), `key_or_hash` MUST be `genesis_bytes` interpreted as a 32-byte SHA-256 hash of a [Genesis Document].
+
+Decoding inverts encoding exactly: passing the `version_number`, `network_name`, and `key_or_hash` produced by this algorithm to the [DID-BTCR2 Identifier Encoding] algorithm MUST reproduce the identical **did:btcr2** identifier string.
 
 
 {% set hide_text = `` %}
--- a/src/algorithms.md
+++ b/src/algorithms.md
@@ -48,7 +48,9 @@ Append `key_or_hash` to the first byte to produce the unencoded data bytes. Its 
 
 [^2]: The "Index" is a Little Endian bit count starting from `0` on the left and ending in `263` or `271` on the right (inclusive).
 
-Encode the unencoded data bytes with Bech32m {{#cite BIP350}} to produce the `method-specific-id`. Use `"k"` as the `hrp` for key-based `key_or_hash` and `"x"` as the `hrp` for [Genesis Document]-based `key_or_hash`.
+Encode the unencoded data bytes with Bech32m [^3] to produce the `method-specific-id`, which MUST be lowercase. Use `"k"` as the `hrp` for key-based `key_or_hash` and `"x"` as the `hrp` for [Genesis Document]-based `key_or_hash`.
+
+[^3]: The `method-specific-id` MUST be conformant to Bech32m {{#cite BIP350}}, which extends Bech32 {{#cite BIP173}}. On encode, pad the final 5-bit group with zero bits; on decode, an incomplete final group MUST be 4 bits or fewer, MUST be all zeros, and is discarded.
 
 Prefix the `method-specific-id` with the string `"did:btcr2:"` to produce the final **did:btcr2** identifier.
 
@@ -61,11 +63,11 @@ Example input:
 * `version_number`: `1`
 * `network_name`: `bitcoin` (`network_value` = `0`)
 * `key_or_hash`: SEC encoded secp256k1 public key
-  `171d59dd2d274011cbb090acc5a168dc98f303790ffce8f54779baed81890c1b00`
+  `0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798`
 
 Example output:
 
-* `did`: `"did:btc1:k1qqt36kwa95n5qywtkzg2e3dpdrwf3ucr0y8le684gaum4mvp3yxpkqqx0845q"`' %}
+* `did`: `"did:btcr2:k1qqp8n0nx0muaewav2ksx99wwsu9swq5mlndjmn3gm9vl9q2mzmup0xqhmkf96"`' %}
 
 {{ ui::show_example_tabs(
   group_id="identifier-encoding-example",
@@ -85,7 +87,7 @@ Parsing a **did:btcr2** identifier produces three values: `version_number`, `net
 
 A **did:btcr2** identifier MUST be processed according to the DID Resolution Algorithm {{#cite DID-RESOLUTION}} to retrieve the **did:btcr2** DID Method-specific ID `method-specific-id`. (I.e., the first 10 characters in the string MUST be exactly `"did:btcr2:"`.)
 
-Decode the `method-specific-id` as a Bech32m encoded string {{#cite BIP350}} to retrieve the unencoded data bytes and `hrp`. Parse the unencoded data bytes according to [Table 2: Unencoded Data Bytes](#unencoded-data-bytes) to retrieve the `btcr2_version`, `network_value` and `genesis_bytes`.
+The `method-specific-id` MUST be lowercase. Decode it as a Bech32m encoded string [^3] to retrieve the unencoded data bytes and the `hrp`. Parse the unencoded data bytes according to [Table 2: Unencoded Data Bytes](#unencoded-data-bytes) to retrieve the `btcr2_version`, `network_value` and `genesis_bytes`.
 
 * `btcr2_version` MUST be `0`. Introduce `version_number` as `btcr2_version + 1`. I.e., `version_number` MUST be returned to the caller as a type that can be represented as the value `1`.
 * `network_value` MUST be one of the values in [Table 1: Network Values](#network-values). `network_name` SHOULD be returned to the caller including additional type information using symbolic names that can be represented as integer values.
--- a/src/references.bib
+++ b/src/references.bib
@@ -18,6 +18,17 @@
   abstract = {Bitcoin Core is the reference implementation of the Bitcoin protocol. It is a full node implementation that validates all transactions and blocks according to the Bitcoin consensus rules.}
 }
 
+@misc{BIP173,
+  title = {Base32 address format for native v0-16 witness outputs},
+  author = {Pieter Wuille, Greg Maxwell},
+  howpublished = {Bitcoin Improvement Proposal 173},
+  year = {2017},
+  month = {March},
+  url = {https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki},
+  note = {BIP-0173},
+  abstract = {This document proposes a checksummed base32 format, "Bech32", and a standard for native segregated witness output addresses using it.}
+}
+
 @misc{BIP321,
   title = {URI Scheme},
   author = {Matt Corallo},


src/algorithms.md
```diff
@@ -32,6 +32,9 @@ The `network_name` value declares which Bitcoin network anchors the identifier.
 | `mutinynet`      | `5`             |
 | Reserved         | `6`..`11`       |
 | Custom networks  | `12`..`14`      |
+| Unassigned       | `15`            |
```
parasyte: This looks like an error by omission. 12-15 are supposed to be the custom networks. From the old spec: https://github.com/dcdpr/did-btcr2/blob/75ae6934458922e70d24d8f3c22db423c7df017f/chapters/Syntax.md?plain=1#L19-L28

src/algorithms.md
```diff
@@ -32,6 +32,9 @@ The `network_name` value declares which Bitcoin network anchors the identifier.
...
+
+`Reserved` and `Unassigned` network values MUST NOT be encoded. A `Custom networks` value MUST NOT be encoded unless the implementation is configured with a network definition for that value.
```
parasyte: I'm not sure about the wording for custom networks. The intention of this change even seems suspicious. The only reason for including custom networks is to allow implementations to use them. It isn't clear what "configuring the implementation with a network definition" means.

The old spec resolves the dilemma by placing the responsibility on the parties who choose to use these network values: https://github.com/dcdpr/did-btcr2/blob/75ae6934458922e70d24d8f3c22db423c7df017f/chapters/Syntax.md?plain=1#L30-L37

In other words, they are free to be used, but no compatibility guarantees are provided by the specification or any implementations that encounter a DID using a custom network. Only the parties involved can agree on how the custom network value is interpreted.

I'm personally not a fan of this kind of extensibility for similar reasons that are raised in #331. (*If it were my call, I would remove it.*)

As for the Chesterton's fence, I believe the reason that custom networks were specified is so that implementations can experiment with their own test networks without violating the spec. (I think the idea was introduced in #14, but justification in the recorded conversation is weak. No idea what was discussed in the call mentioned.) I can see no way for these custom networks to ever be promoted to a fully supported network in a future version of the specification with the same value. They are a permanent but small sandbox where collisions are expected.

To me, custom networks do not make sense to specify. It's just a "blessed reservation" where all bets are off. And unlike the specified networks, custom networks **CANNOT** be used interoperably across versions because the specification reserves the right to reallocate those 4 values for another purpose.

<details><summary>Versioning scheme details</summary>

By specifying that the `btcr2_version` nibble MUST be `0`, it is declaring that all other versions cannot be handled by an implementation that only supports this version of the spec. In turn, a successor version can choose to do anything it wants with the encoding and it will not cause any conflicts. Comments in #14 show that this was intentional.
</details>

As far as *this* particular change goes, I might rephrase it so that custom networks can be used for any purpose without limitation or guarantee. Similar to what appears in the old spec. There is no need to specify what "configuring a network definition" means, since all beta are off anyway.

src/algorithms.md
```diff
@@ -85,16 +88,22 @@ Parsing a **did:btcr2** identifier produces three values: `version_number`, `net
...
 
-Decode the `method-specific-id` as a Bech32m encoded string {{#cite BIP350}} to retrieve the unencoded data bytes and `hrp`. Parse the unencoded data bytes according to [Table 2: Unencoded Data Bytes](#unencoded-data-bytes) to retrieve the `btcr2_version`, `network_value` and `genesis_bytes`.
+The `method-specific-id` MUST be lowercase. Decode it as a Bech32m encoded string {{#cite BIP350}} to retrieve the 5-bit data groups and `hrp`, then convert the 5-bit groups back to 8-bit bytes using the conversion defined by BIP173 {{#cite BIP173}} to retrieve the unencoded data bytes. Following the BIP173 decoding rule, the incomplete group of padding bits at the end MUST be 4 bits or fewer, MUST be all zero bits, and is discarded; a `method-specific-id` that violates either condition MUST be rejected. Parse the unencoded data bytes according to [Table 2: Unencoded Data Bytes](#unencoded-data-bytes) to retrieve the `btcr2_version`, `network_value` and `genesis_bytes`.
```
parasyte: It is the case that BIP-173 requires 5-bit to 8-bit grouping and zero padding bits. Why do we need to explicitly say that?

We don't talk about the bitwise operations like Xor and rotations that go into computing a SHA-256 hash. Why should be talk about the grouping and padding bits that go into a BIP-173 conversion?

src/algorithms.md
```diff
@@ -85,16 +88,22 @@ Parsing a **did:btcr2** identifier produces three values: `version_number`, `net
...
 
 * `btcr2_version` MUST be `0`. Introduce `version_number` as `btcr2_version + 1`. I.e., `version_number` MUST be returned to the caller as a type that can be represented as the value `1`.
-* `network_value` MUST be one of the values in [Table 1: Network Values](#network-values). `network_name` SHOULD be returned to the caller including additional type information using symbolic names that can be represented as integer values.
+* `network_value` MUST be handled according to its row in [Table 1: Network Values](#network-values):
+  * A named network value (`0`..`5`) maps to its `network_name`. `network_name` SHOULD be returned to the caller including additional type information using symbolic names that can be represented as integer values.
+  * A `Reserved` value (`6`..`11`) MUST be rejected until this specification assigns it a `network_name`.
+  * A `Custom networks` value (`12`..`14`) MUST be rejected unless the implementation is configured with a network definition for that value, in which case `network_value` maps to the configured network's `network_name`. How implementations agree on and configure custom network definitions is out of scope for this specification.
```
parasyte: Again, "configuring a network definition" is itself not defined. It can't be declared "out of scope" because the acceptance of custom network values is conditional on an implementation meeting the requirement to configure a custom network definition.

It would be better if this said something like "SHOULD be rejected when the implementation does not support a custom network for that value" where "supporting a custom network" is implementation-defined. Whether or not a custom network is rejected doesn't matter. The "MUST" is too strong.

parasyte (commented):
Lots to think about, here.

src/algorithms.md
jintekc: Good catch. Will change to align with old spec.

src/algorithms.md
jintekc: All of that is fair push back. I agree it makes the most sense to adopt the old spec's framing.

src/algorithms.md
jintekc: Point taken in prior comment re: old-spec direction.

src/algorithms.md
jintekc: I happen to agree with you here. However, @danpape and I thought it might make sense to call it out explicitly in this limited context given the possibility for collisions and errors IF an implementation created their own encoder or used a bech32m library that leaves the byte conversion to the caller.

That said, the verbosity isn't necessary. Maybe something simpler would suffice:

The method-specific-id MUST be conformant to Bech32m {{#cite BIP350}} and to the base32 conversion rules defined by BIP173 {{#cite BIP173}}.

parasyte: Those concerns can be covered by the test suite by asserting that invalid DIDs return the appropriate parsing errors.

danpape: The problem with BIP173 is that it describes two different things. First, the bech32 algorithm itself, and second, the SegWit address format, which uses bech32. It is the description of the SegWit address format which specifies the zero-padding that should be used and checked when encoding and decoding the witness program. The bech32 algorithm itself works with or without the padding. We could say that our DID identifier needs to be encoded and decoded using the SegWit address algorithm, but that seems weird because we have nothing to do with SegWit. So we should mention the padding needed for encoding and decoding our DIDs, but I guess I see how it makes this text even wordier. Perhaps instead of including it here, maybe we put it in a separate small section of this algorithms chapter, or a footnote or something like that? A couple extra tests would be good too. I could work on this and push something up in a bit?

parasyte: I suppose BIP-173 is the wrong reference for these encoding details, if that only applies to SegWit addresses. We already cite BIP-350 for the Bech32m encoding, which delegates most of the details to BIP-173, with the exception of the altered checksum constant. So, we don't need any of the Bech32 references from BIP-173, AFAICT.

If BIP-173 SegWit address padding rules do not apply to did-btcr2 encoding, then I don't see a reason to cite BIP-173. We can maintain our own padding rules and allow BIP-350 (and its delegation to BIP-173) to define the 5-to-8 conversion.

I'm OK with a wordier description if we need to define padding requirements because other specifications do not. I just don't want to cite a specification that doesn't do what it's intended to.

danpape: Good point. We probably should point to BIP350, especially since all common bech32 libraries (at least the ones the did:btcr2 implementations we know about use) default to the `bech32m` variant. That said, as @jintekc found while testing, if you don't set the padding to zero sometimes it can come out random, so it is a good idea if we at least mention it.

jintekc: To be precise about what each BIP is doing:

  * BIP-350 changes only the checksum constant ([BIP-350 L38](https://github.com/bitcoin/bips/blob/fe2d6caf85f976794095ac5c60c11ccee81e50b2/bip-0350.mediawiki?plain=1#L38)): "All other aspects of Bech32 remain unchanged" ([BIP-350 L65](https://github.com/bitcoin/bips/blob/fe2d6caf85f976794095ac5c60c11ccee81e50b2/bip-0350.mediawiki?plain=1#L65)).
  * The Bech32 section of BIP-173 never touches bytes. Its data part is defined as characters ([BIP-173 L89](https://github.com/bitcoin/bips/blob/fe2d6caf85f976794095ac5c60c11ccee81e50b2/bip-0173.mediawiki?plain=1#L89)) handled as "a list of integers representing the characters" ([BIP-173 L132](https://github.com/bitcoin/bips/blob/fe2d6caf85f976794095ac5c60c11ccee81e50b2/bip-0173.mediawiki?plain=1#L132)). No byte conversion appears anywhere in that section.
* The 8-bit-to-5-bit mapping and its zero padding are defined in exactly one place, the segwit address section: "Re-arrange those bits into groups of 5, and pad with zeroes at the end if needed" on encode ([BIP-173 L214-L215](https://github.com/bitcoin/bips/blob/fe2d6caf85f976794095ac5c60c11ccee81e50b2/bip-0173.mediawiki?plain=1#L214-L215)), and "Any incomplete group at the end MUST be 4 bits or less, MUST be all zeroes, and is discarded" on decode ([BIP-173 L225](https://github.com/bitcoin/bips/blob/fe2d6caf85f976794095ac5c60c11ccee81e50b2/bip-0173.mediawiki?plain=1#L225)).

So pointing at BIP-350 alone doesn't cover the padding and wouldn't cover the non-deterministic DID creation demoed in #330.

And pointing at BIP-173 alone doesn't include the right checksum, since the bech32m constant only exists in BIP-350.

I think they're both useful and important in a footnote.

danpape (dismissed):

src/algorithms.md
parasyte: That isn't exactly filling me with confidence. :cry: Encoding and decoding is only specified for SegWit addresses. Does it mean that Bech32m encoding is undefined for anything other than SegWit addresses? On the bright side, the 5-to-8 conversion and zero padding bits are the obvious choice for any base32 encoding.

The "list of integers" for the data part only has integers in the range of 0 - 31 (it's a base32 encoding, after all) and is defined as an input for checksum verification. It's just a fancy way of saying the checksum operates over the encoded characters, not the decoded bytes.

> So pointing at BIP-350 alone doesn't cover the padding ...

BIP-173 does, but only for SegWit addresses. BIP-350 can't be implemented without first implementing the majority of BIP-173. The SeqWit-specific encoding details notwithstanding. Reading BIP-350 as if it is a complete, standalone specification is not a meaningful interpretation.

> ... and wouldn't cover the non-deterministic DID creation demoed in #330.

FWIW, implementations will only have non-deterministic encoding (non-zero padding bits) if they reuse the bit rotation state without clearing it to zero when there is a wrap-around. I wrote such a base32 codec recently that doesn't reuse rotation state. It pre-allocates a zeroed buffer and updates those bytes directly. The zeroed buffer ensures that all padding bits are zero.

> I think they're both useful and important in a footnote.

If pointing at both is a marked improvement, then by all means.

I'm resisting because even with two references, using Bech32m is *still* underspecified. I get it, the DID wants to resemble a Bitcoin address. Using base64url-nopad would have solved all of these problems. We already reference it for update hashes (originally proposed in #58, fixed by #122, tracked in #219, and improved by #315). Or, alternatively, tie up the leaks with Bech32m and use it consistently for all raw byte encoding. Or, maybe go back to base58-btc as proposed in #301 :joy:. There are various encoding methods in use in the specification, and Bech32m seems to be the hardest to clearly define.

parasyte (dismissed):
So you don't want to specify the padding after all? Or point it out when referring to BIP-173? Or the test vectors in BIP-350, for that matter?

src/algorithms.md
danpape: Agreed that Bech32m outside SegWit is underspecified. The bech32 section of BIP-173 only defines 5-bit values 0–31 -> alphabet and never touches bytes. Any application encoding common 8-bit data has to define its own 8->5 repacking and what happens to the leftover bits, and BIP-173 only does that inside the SegWit address section. We're a different application, so those rules aren't ours by reference.

So, I think that means you're right that we shouldn't cite BIP-173 for it. Better to state the rule ourselves: on encode, pad the final group with zero bits; on decode, an incomplete final group MUST be 4 bits or fewer, MUST be all zeros, and is discarded. Then we cite BIP-350 for Bech32m, and either drop BIP-173, or leave it in as a reference since 350 extends it.

As for why bech32 at all: the checksum. We expected the DID strings to get copy/pasted and occasionally typed by hand, and bech32m gives real error detection for that, BIP-173 quantifies it. base64url-nopad is simpler and deterministic but gives none. The extra complication and checksum is also why we don't use it for everything; update hashes never cross a human boundary, so the checksum isn't worth the cost there.

parasyte: Cross-posting my response from the Signal group chat (with addendums).

Some good points. I don’t think the checksum adds any value, but that’s just me. It makes decoding failures determinable sometimes. (Checksum matches are probabilistic, not cryptographically secure.)

An invalid Bech32m DID with a valid checksum is the same as an invalid Base64 DID. rxgrant gets upset with me when I bring up pragmatic details like this. The presence of a checksum does not prevent things from going wrong. It just adds a new state for implementations to deal with.

An error-correcting code, on the other hand (like QR codes have) would be more convincing for the hand-written use case. But again, *that's just, like, my opinion, man*.

danpape: Bech32 actually is an error correcting code. It's a BCH code, and BIP173's appendix compares that to the Reed-Solomon codes QR uses. The BIPs, however, suggest not using the correction part. BIP-173 has an "Error correction" section and BIP350 has an "Error locating" section, and both say the same thing: you can correct a few errors, but correcting erodes detection, so implementations SHOULD NOT do more than point at where the error probably is. Sipa (bip author) wrote an error locating decoder and a demo page for it.

If we had no checksum for the DID identifier (if we used base64url-nopad), then any k1 did:btcr2 has no lookup that can fail, since the document is derived straight from the key in the string. So a typo wouldn't give you a "not found". The changed bytes could still land on a valid secp256k1 curve point, and you'd get a valid DID for a document that isn't yours with nothing to catch it. The checksum is what helps stop that from ever getting that far, and it's why base64url-nopad wouldn't work for us. And as the bip suggests, while the error detection is good, but we shouldn't error correct our DID identifiers. If we correct badly, we could again get a valid string that points somewhere else.

jintekc:
made a final edit to the footnote that aligns with the suggestion made by @danpape 

> \[^3]: The method-specific-id MUST be conformant to Bech32m {{#cite BIP350}}, which extends Bech32 {{#cite BIP173}}. On encode, pad the final 5-bit group with zero bits; on decode, an incomplete final group MUST be 4 bits or fewer, MUST be all zeros, and is discarded.

danpape (approved):

parasyte (approved):

src/algorithms.md
parasyte: It is surprising to see that the error-correcting code is not recommended to be used for error correction, but not necessarily surprising that the checksum *is* an ECC.

> If we had no checksum for the DID identifier (if we used base64url-nopad), then any k1 did:btcr2 has no lookup that can fail ...

Which is fine, because of cryptographic strength (which I'll get back to momentarily).

> So a typo wouldn't give you a "not found".

That's the "new state to deal with" that I mentioned before. An error here is not saying the DID is "not found". It is saying something else like, "the DID is corrupted". Which is a very different interpretation and has a very different mechanism from being "found or not found". A "not found" `k1` did:btcr2 DID is one that doesn't have a corresponding private key. (Cryptographic strength, I'm getting there.)

> The changed bytes could still land on a valid secp256k1 curve point ...

There it is! Now we can talk about cryptographic strength. Finding an arbitrary collision between _any two keys_ in an *n*-bit cryptographic key-space requires `2^(n/2)` brute-force attempts for a \~50% probability of finding any such collision (the birthday problem). At `n = 256`, `1^128` is a satisfyingly large enough number that a birthday attack is infeasible.

However, we are not concerned with any two arbitrary collisions. The attack you are implying is closer to a preimage attack, but more like a near-miss than a collision. For any 1 bit difference, a brute force attack is expected to take `2^(n - 1)` attempts. And targeting a *specific* single bit requires far more.

The probability that a valid but unexpected secp256k1 curve point is reached and *has a known private key* for any small number of bit flips is astronomically unlikely. I have faith in the power of large numbers (I would not use 256-bit cryptography otherwise) but I have no faith in (computationally) small checksum probabilities being a strong protection.

> The checksum is what helps stop that from ever getting that far ...

It does with a caveat. The probability of the checksum passing with typos is on the order of 1 per billion (~1 in `2^30`) according to BIP-173. That's more than good enough for human typing, but inadequate for error detection at a reasonable scale. (Although, I acknowledge that *other* checksums are already prevalent in every transmission protocol anyone would reasonably use. The Bech32m checksum is redundant for the at-scale case.)

I've come to discover that users are expected to have a large number of DIDs; one for each pairwise relationship with another identity, one for each group they are a member of, one for each device they own, and any combination of those. For a socially active user, this could easily be multiple millions or billions of DIDs. That's already in the ballpark for 1 in a billion to be significant, and that's just a single user.

And "stops that from ever getting that far" isn't really useful, either. The moment you try to resolve a DID with typos, you'll find nothing (established with cryptographic strength). You know it is truly "not found", and not just "got corrupted in transit".

The Bech32m checksum is truly only useful for a fairly rare human-oriented problem (typos) and is otherwise a waste of space and computation.

> And as the bip suggests, while the error detection is good, but we shouldn't error correct our DID identifiers. If we correct badly, we could again get a valid string that points somewhere else.

It's still surprising that error correction is not recommended. Getting a valid string that points somewhere else is not surprising at all, and again, it's perfectly fine thanks to large numbers.

----

It's not my choice to make. There are reasons for using it. I disagree with those reasons. And as a result, the specification has at least three different ways to encode raw bytes. I'm leaning toward the Zen approach, "this is not my problem" and moving on. But, I'm glad the use of Bech32m for DID encoding is sorted out.

Issue #349, opened 2026-08-21
Decide which DID Core version we conform to when our two normative targets disagree


danpape:
[conformance.md:14](https://github.com/dcdpr/did-btcr2/blob/df2748f/src/conformance.md?plain=1#L14) requires conformance to two documents at once:

> Implementations MUST be conformant to all normative statements in Decentralized Identifiers v1.1 and DID Resolution v1.

Those two disagree about which DID Core applies. DID Resolution v1 normatively references DID Core **v1.0**, the 2022 Recommendation:

> [DID-CORE] Decentralized Identifiers (DIDs) v1.0. W3C Recommendation. `https://www.w3.org/TR/did-core/`

So we mandate conformance to a resolution specification that is itself written against an older DID Core than the one we also mandate. Worth noting that both documents are still Candidate Recommendations, DID Core v1.1 dated 05 March 2026, DID Resolution v1 dated 06 August 2026, so what we require can still change under us.

I figure we ought to still require DID Core v1.1, and mention what happens where the two disagree. Something like this, added to the conformance section:

> Where a normative statement in DID Resolution v1 relies on Decentralized Identifiers v1.0, the corresponding statement in Decentralized Identifiers v1.1 applies to **did:btcr2** identifiers and DID documents.

That matches what we already do in practice and costs one sentence. Dropping the DID Core dependency down to v1.0 is probably not possible, since we would lose `application/did`, and probably other stuff I haven't noticed.

parasyte:
The easiest thing to do is rollback to DID-CORE v1.0. Which properties of v1.1 were appealing enough to update it?

danpape:
Mmm, good point. I assume we went with 1.1 since when we started did:btcr2 the DID working group was already working on getting DID Core 1.0 updated and we thought it would get finalized more quickly. 

As for the big changes between 1.0 and 1.1, they have a list here: https://www.w3.org/TR/did-1.1/#revision-history 

Besides just the media type change, if we stopped depending on 1.1, even just for a while until they finalized, we'd have to redo all the current btcr2 test vectors since the `@context` value would have to change. DID core 1.1 and our own [data_structures.md:23](https://github.com/dcdpr/did-btcr2/blob/df2748fb4f/src/data-structures.md?plain=1#L23) say something similar to 

> `@context` array MUST contain "https://www.w3.org/ns/did/v1.1". 

If we went with 1.0, that spec says 

> `@context` MUST be the JSON String https://www.w3.org/ns/did/v1

[...]

PR #350, opened 2026-08-22
Tweak normative references

### LINKED ISSUE #348: Pin the two W3C citations the same way, and fix the DID Core date
`references.bib` carries our two W3C normative references, pinned by opposite strategies:

* [`DID-CORE`](https://github.com/dcdpr/did-btcr2/blob/df2748f/src/references.bib?plain=1#L98-L106) uses the moving "latest version" URL, `https://www.w3.org/TR/did-1.1/`.
* [`DID-RESOLUTION`](https://github.com/dcdpr/did-btcr2/blob/df2748f/src/references.bib?plain=1#L113-L121) uses a frozen dated snapshot, `https://www.w3.org/TR/2026/CR-did-resolution-1.0-20260806/`.

Two smaller inconsistencies sit in the same six lines:

* `DID-CORE` says `year = {2025}, month = {September}`. DID Core v1.1 is a W3C Candidate Recommendation Snapshot dated **05 March 2026**.
* `DID-RESOLUTION` records its status in `howpublished` ("W3C Candidate Recommendation Snapshot, 06 August 2026") while `DID-CORE` says "Online".

Two ways to fix this:

**Track**: both use latest-version URLs (`https://www.w3.org/TR/did-1.1/` and `https://www.w3.org/TR/did-resolution-1.0/`). The citation stays correct as each document advances toward Recommendation, and there is nothing to remember. The cost is that the text we cite can change underneath us, which is what produced the renumbering that #339 dealt with.

**Freeze**: both use dated snapshots. Reproducible, at the cost of a deliberate re-pin at every transition. Nothing currently tracks that, and DID Resolution will not advance to Recommendation earlier than 03 September 2026, with DID Core v1.1 pending as well. A dated URL keeps resolving forever, so a stale pin does not break, it just quietly cites a superseded document.

At this point, I think I'm leaning toward tracking the latest versions, because one reason to freeze the DID resolution reference was the fragile section numbers as they added new stuff, and our `errors.md` now cites the spec's Errors section by name rather than by number. If we take that route, citing by name everywhere should be the stated rule, not an accident of one edit.

Either way, we should correct the `DID-CORE` date and make `howpublished` consistent across the two.

danpape:
Updates references for DID core and DID resolution specs. Also adds a statement regarding our temporary dependency on both 1.0 and 1.1 versions of DID core spec.

Fixes #348 and #349

### DIFF
--- a/src/conformance.md
+++ b/src/conformance.md
@@ -12,3 +12,5 @@ The key words MAY, MUST, MUST NOT, RECOMMENDED, SHOULD, and SHOULD NOT in this d
 Interoperability of implementations of the **did:btcr2** DID method is tested by evaluating an implementation's ability to create, resolve, update, and deactivate, **did:btcr2** identifiers and DID documents that conform to this specification. Interoperability for producers and consumers of **did:btcr2** identifiers and DID documents is provided by ensuring the DIDs and DID documents conform.
 
 Implementations MUST be conformant to all normative statements in Decentralized Identifiers v1.1 {{#cite DID-CORE}} and DID Resolution v1 {{#cite DID-RESOLUTION}}.
+
+DID Resolution v1 normatively references Decentralized Identifiers v1.0. Where a normative statement in DID Resolution v1 relies on Decentralized Identifiers v1.0, the corresponding statement in Decentralized Identifiers v1.1 applies to **did:btcr2** identifiers and DID documents.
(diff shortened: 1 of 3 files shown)


parasyte (approved):
Seems reasonable to me. Hopefully the cross-referencing between Decentralized Identifiers v1.0 and DID Resolution v1 does not end up causing unforeseen issues.

jintekc (approved):
LGTM

Issue #341, opened 2026-08-11
Define SMT leaf values for both `nonce` modes


jintekc:
## Problem

`nonce` is OPTIONAL in an SMT Proof ([data-structures.md:254](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/data-structures.md?plain=1#L254)). However, both arms of the leaf reconstruction ([algorithms.md:169-173](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/algorithms.md?plain=1#L169-L173)) hash `proof.nonce`. Therefore a proof without a `nonce` always fails verification. The normative text defines no leaf value for an occupied index without a `nonce`.

The only no-nonce leaf values in the spec are in the aggregation walkthrough. That walkthrough is a non-normative example ([aggregate-beacons.md:12](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/beacons/aggregate-beacons.md?plain=1#L12), [conformance.md:8](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/conformance.md?plain=1#L8)). Both of its entries are stale. No verifier arm reconstructs the bare `json_document_hash(update)` at [:81](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/beacons/aggregate-beacons.md?plain=1#L81). The literal `0` at [:84](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/beacons/aggregate-beacons.md?plain=1#L84) differs from the empty leaf value, which is `hash(0 + 0)` ([optimized-smt.md:124](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/appendix/optimized-smt.md?plain=1#L124)).

## Proposed Solution

Keep `nonce` OPTIONAL. Define the leaf value for every combination of the two optional fields. Each index in each signal has its own combination:

```rust
match (proof.nonce, proof.updateId) {
  (Some(nonce), Some(updateId)) => hash(concat(hash(nonce), updateId)),  // update, private
  (Some(nonce), None)           => hash(hash(nonce)),                    // no update, private
  (None,        Some(updateId)) => updateId,                             // update, public
  (None,        None)           => cachedZero[0],                        // empty index: hash(concat(0, 0))
}
```

[...]

Issue #323, opened 2026-07-21
Clarify versionId/versionTime resolution semantics


jintekc:
Three gaps in the definition of `resolutionOptions` can cause conformant resolvers to return different results for the same query.

### Problem

The spec defines no result of `resolve` for a requested `versionId` that is larger than the last version of the DID. For example, the request specifies `versionId` 5 and the history ends at version 3. If one resolver returns version 3 and a second resolver raises an error, the result is a consensus split. A substitution without an error makes a verifier validate against the wrong document.

### Proposed Solution

Raise an error. Map the error to the `NOT_FOUND` error of DID Resolution, or to a method-specific code from the `errors.md` registry work.

### Problem

[resolve.md:153](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L153) compares "the tuple's block time" but does not name the timestamp field. A Bitcoin block carries two timestamps: the header timestamp and the median-time-past (MTP). The header timestamp can decrease from one block to the next. A miner can set the header timestamp up to approximately 2 hours in the future. MTP is monotonic and non-decreasing.

On a typical block, the two timestamps differ by tens of minutes. Two resolvers that compare different fields select different versions near a boundary. [data-structures.md:285](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/data-structures.md?plain=1#L285) defines `versionTime` as an exclusive boundary ("valid ... before the specified versionTime"). [resolve.md:153](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L153) uses a "more recent" test, which is inclusive at equality. One chapter applies an update whose block time equals `versionTime`. The other chapter excludes that update.

### Proposed Solution

Use the median-time-past of the block that contains the Beacon Signal transaction as the comparison basis. The bitcoind RPCs `getblockheader` and `getblock` return this value directly. Esplora `GET /block/:hash` returns it as `mediantime`. An implementation reads the value from one of these sources. An implementation without such a source computes the median of the header timestamps of the block and its 10 ancestors.

The comparison is exact. It uses no tolerance and no grace window. MTP is an approximate time. Every resolver reads the same MTP value for the same block. Determinism is more important than accuracy here.

[...]

PR #351, opened 2026-08-22
Pin the update `@context`, harden the resolve-path proof checks, regenerate the examples

### LINKED ISSUE #259: Regenerate all examples in the spec that involve Proof Value. The migration to btcr2 means these are no longer accurate.
This should likely be done as part of a broader review of examples throughout the spec

### LINKED ISSUE #324: Harden the resolve-path proof checks
The Check `update.proof` step in ([resolve.md:194-200](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L194-L200)) currently mandates only three checks:
1. locate the verification method
2. require `capabilityInvocation` membership
3. verify the BIP340 signature

The full ZCAP invocation is a MAY. This leaves four gaps where conformant resolvers diverge or where the data model permits documents the algorithm cannot honor. 

1. `expires` MUST NOT be evaluated on the resolve path
* The resolve algorithm never mentions proof `expires`, but [resolve.md:200](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L200) delegates verification to "a BIP340 Cryptosuite instance", and generic Data Integrity libraries evaluate `expires` against the current wall clock in their envelope-verification layer. A library-based resolver therefore rejects (fatally, per [errors.md](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/errors.md)) an update that a bare signature-check resolver accepts, and even disagrees with its own past self once the timestamp passes. Wall clock is not an input to resolution: resolution must remain a pure function of the DID, the resolution options, and the on-chain plus sidecar data.
* Proposal: add to the "Check update.proof" section: "Resolvers MUST NOT evaluate the proof's `expires` (or `created`) values on the resolve path; these are construction-time metadata." Implementations using generic Data Integrity libraries must disable the envelope expiry check.

2. Make the three fixed-value checks MUSTs in the resolve-path
* [data-structures.md:167-170](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/data-structures.md?plain=1#L167-L170) fixes `proofPurpose` (`"capabilityInvocation"`), `capabilityAction` (`"Write"`), and `capability` (`urn:zcap:root:${encodeURIComponent(did)}`) at construction, but the resolve path never re-checks them; only the OPTIONAL (MAY) ZCAP invocation path does. A validly-signed update whose `capability` names a different DID, or whose action is not `"Write"`, is accepted by mandatory-checks-only resolvers and rejected by ZCAP-invoking resolvers today, with no attacker required.
* Proposal: add three MUSTs to the "Check update.proof" section with failures raising `INVALID_DID_UPDATE`. These are one-line equality checks over signed bytes:
  * `update.proof.proofPurpose` MUST equal `"capabilityInvocation"`
  * `update.proof.capabilityAction` MUST equal `"Write"`
  * `update.proof.capability` MUST equal `urn:zcap:root:${encodeURIComponent(did)}` for the DID being resolved

3. Restrict `capabilityInvocation` to verification-method references
* [data-structures.md:35-36](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/data-structures.md?plain=1#L35-L36) permits embedded verification-method objects in `capabilityInvocation`, but an embedded verification method can never satisfy either of the conditions in the [resolve.md:198](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L198) path: (a) locate the key only in the top-level `verificationMethod` array; (b) test whether `capabilityInvocation` contains the proof's `verificationMethod` id string. Thus, a document built that way is "updatable" per the data model but permanently un-updatable in practice: every conformant resolver rejects every update it signs.

* Proposal: in the DID Document data structure, require that `capabilityInvocation` entries in did:btcr2 documents be references to verification methods declared in the top-level `verificationMethod` array; embedded objects are not permitted. (All spec examples already conform.)

4. matching rule for relative vs absolute references
* DID Core permits relative DID URLs in documents (e.g. `#key-1`) while proofs carry absolute DID URLs (`did:btcr2:…#key-1`). Plain string equality quietly fails on that pair, so a valid update is rejected by a naive resolver and accepted by a normalizing one. 
* Proposal 1: forbid relative references in did:btcr2 documents: all verification-method ids and references MUST be absolute. Stricter and simpler; matches every current example; rejects documents that DID Core itself considers valid.
* Proposal 2:  normalize; resolve relative references against the document `id` (per DID Core) before exact comparison. DID Core-correct; accepts documents produced by common tooling; one extra step in resolvers.

Welcoming feedback and discussion on these proposals and the open question.

### LINKED ISSUE #325: Pin the update `@context` array
Two issues with the current update `@context` array defined in [`data-structures.md:97-101`](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/data-structures.md?plain=1#L97-L101): the content and order of the URLs must be finalized and how it gets verified via cryptosuite must be settled.

1. The update `@context` MUST be exactly this array, in this order:

```json
[
  "https://w3id.org/json-ld-patch/v1",
  "https://w3id.org/zcap/v1",
  "https://w3id.org/security/data-integrity/v2",
  "https://btcr2.dev/context/v1"
]
```

`security/data-integrity/v2` is the context that defines the terms the proof actually uses (`DataIntegrityProof`, `cryptosuite`, `proofValue`); `security/v2` is the older security vocabulary and leaves those terms undefined for JSON-LD processors, so it is dropped. `btcr2.dev/context/v1` is required; the sidecar example's omission is a bug.

The proposed order places the method context (`btcr2.dev/context/v1`) last, following the common JSON-LD most-specific-last convention so the method context wins term conflicts, with the W3C contexts first.

2. verification-time `@context` comparison should be exact match

The bip340-jcs cryptosuite's generic verification allows the proof's `@context` to be a prefix of the secured document's, replacing the document's array with the proof's before canonicalization. For did:btcr2 updates that tolerance is unreachable: an update's identity is its JCS hash, `@context` included, so any post-signing context change alters the update hash and fails the on-chain announcement lookup ([resolve.md:129-131](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L129-L131)) with `MISSING_UPDATE_DATA` before proof verification ever runs. The spec should therefore state, as a method-level constraint (a tightening of the cryptosuite, not a deviation from it): the proof's `@context` MUST equal the update's `@context`, and both MUST equal the array above; resolvers MUST reject any mismatch with `INVALID_DID_UPDATE`. When the arrays are required to be equal, the cryptosuite's prefix check degenerates to equality.

jintekc:
Bundling issues #325 #324 and #259.

- **#325**: pins the update `@context` to an exact array in a fixed order, and makes the resolve path reject an update whose context, or whose proof context, does not match.
- **#324**: hardens the resolve-path proof checks. The three values construction fixes are re-checked at resolution, the proof's validity window is decided by the announcing block rather than by when resolution runs, embedded verification methods work on both paths, and relative DID URLs are resolved against the document `id` before comparison.
- **#259**: regenerates `src/example-data` from committed seeds with `bin/gen-examples.ts`, so a re-run reproduces the same bytes and `--check` verifies the files without writing.

The SMT vector needs a re-run when #341 lands.

Fixes #325
Fixes #324
Fixes #259

### DIFF
--- a/bin/gen-examples.ts
+++ b/bin/gen-examples.ts
@@ -1,0 +1,623 @@
+#!/usr/bin/env bun
+/**
+ * Regenerates every file in src/example-data from the seeds in this file.
+ *
+ *   bun bin/gen-examples.ts           write the corpus
+ *   bun bin/gen-examples.ts --check   verify the committed corpus, write nothing
+ *
+ * Nothing in the corpus is hand-written, and every step the reference
+ * implementation covers is performed by the reference implementation. The
+ * @did-btcr2/api SDK is the entry point and does as much as it reaches:
+ * identifiers, key pairs, multikeys, Data Integrity proofs, and the DID document
+ * classes it re-exports. The packages under it supply what it does not surface,
+ * listed at the import block. Every proof is verified and every hash is recomputed
+ * before anything is written, so a run that succeeds has verified the corpus
+ * rather than just serialized it. Where the implementation has not caught up to
+ * the specification, the specification wins and the difference is marked GAP.
+ * There are four. GAP 3 is one theme, the order of JSON properties, and it
+ * appears at three sites. JCS sorts member names, so property order never
+ * changes a hash; the reorder only makes the committed files read in the order
+ * the specification lists.
+ *
+ * Runs are reproducible: secret keys are SHA-256 of their seed string and BIP340
+ * aux_rand is SHA-256 of a per-signature label. Changing either changes every
+ * proofValue in the corpus, so leave them pinned.
+ */
+
+import { existsSync, readFileSync, writeFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+// The SDK covers identifiers, key pairs, multikeys, and Data Integrity proofs, and
+// re-exports the DID document classes. The imports below it are used only for the
+// parts the SDK does not cover: the update operation (`DidMethodApi.update` and
+// `UpdateBuilder.execute` both broadcast a Bitcoin transaction, and this corpus is
+// not anchored to a chain), Beacon Services, root capability derivation, hashing and
+// patching, and the Sparse Merkle Tree, which the SDK does not expose.
+import type { DidVerificationMethod, PatchOperation } from '@did-btcr2/api';
+import { createApi, DidDocument, GenesisDocument, IdentifierTypes } from '@did-btcr2/api';
+import { getNetwork } from '@did-btcr2/bitcoin';
+import { canonicalHash, canonicalize, JSONPatch } from '@did-btcr2/common';
+import type { Signer, SchnorrKeyPair, SigningScheme } from '@did-btcr2/keypair';
+import type { BeaconService } from '@did-btcr2/method';
+import { Appendix, BeaconUtils, Updater } from '@did-btcr2/method';
+import {
+  BTCR2MerkleTree,
+  didToIndex,
+  hashToBase64Url,
+  inclusionLeafHash,
+  verifySerializedProof,
+} from '@did-btcr2/smt';
+import { schnorr } from '@noble/curves/secp256k1';
+import { sha256 } from '@noble/hashes/sha2';
+import { utf8ToBytes } from '@noble/hashes/utils';
+import { p2pkh, p2tr, p2wpkh } from '@scure/btc-signer';
+
+const EXAMPLES = new URL('../src/example-data/', import.meta.url).pathname;
+
+/** No Bitcoin or CAS config: every sub-facade this script touches is offline. */
+const api = createApi();
+
+/** Every identifier in the corpus is on mutinynet (network_value 5, Table 1). */
+const NETWORK = 'mutinynet';
+const BTCR2_VERSION_NUMBER = 1;
+
+/** Placeholder identifier of a Genesis Document. */
+const PLACEHOLDER = 'did:btcr2:_';
+
+/**
+ * Secret key seeds. The key material is worthless by construction: anyone can
+ * recompute these keys from the strings, which is the point for an example corpus.
+ */
+const SEEDS = {
+  key0: 'did:btcr2 example corpus / controller / key-0',
+  key1: 'did:btcr2 example corpus / controller / key-1',
+  casAggregator: 'did:btcr2 example corpus / CAS aggregator',
+  smtAggregator: 'did:btcr2 example corpus / SMT aggregator',
+  keyBased: 'did:btcr2 example corpus / key-based identifier',
+  cohortA: 'did:btcr2 example corpus / cohort member A',
+  cohortB: 'did:btcr2 example corpus / cohort member B',
+} as const;
+
+/** SMT nonces, one per index per signal. */
+const NONCE_SEEDS = {
+  primary: 'did:btcr2 example corpus / SMT nonce / controller',
+  cohortA: 'did:btcr2 example corpus / SMT nonce / cohort member A',
+  cohortB: 'did:btcr2 example corpus / SMT nonce / cohort member B',
+} as const;
+
+/**
+ * Bitcoin block metadata. The corpus is not anchored to a real chain, so these
+ * illustrate the shape of resolution output rather than describing real blocks.
+ */
+const LAST_UPDATE_TIME = '2025-01-06T16:23:10Z';
+const RESOLUTION_TIME = '2025-01-07T00:00:00Z';
+const CONFIRMATIONS = 12;
+
+// ---------------------------------------------------------------------------
+// Primitives
+// ---------------------------------------------------------------------------
+
+type Json = any;
+
+const seedKey = (seed: string) => api.crypto.keypair.fromSecret(sha256(utf8ToBytes(seed)));
+
+/** The bytes an operation hashes and signs over: the JCS form of a document. */
+const canonicalBytes = (document: Json) => utf8ToBytes(canonicalize(document));
+
+/**
+ * Fails unless two objects are the same JSON document. JCS sorts keys, so this
+ * compares content and ignores key order, which is what separates a corpus that
+ * disagrees with the implementation from one that only serializes differently.
+ */
+function agree(what: string, produced: Json, emitted: Json): void {
+  if (canonicalize(produced) !== canonicalize(emitted)) {
+    throw new Error(
+      `${what}: the implementation and the corpus disagree\n`
+      + `  implementation: ${canonicalize(produced)}\n`
+      + `  corpus:         ${canonicalize(emitted)}`,
+    );
+  }
+}
+
+/** Fills a `{{variable}}` template and parses the result, as the operations do. */
+function render(templateFile: string, vars: Record<string, string>): Json {
+  const template = readFileSync(join(EXAMPLES, templateFile), 'utf8');
+  const filled = template.replace(/\{\{\s*([\w-]+)\s*\}\}/g, (_, name: string) => {
+    if (!(name in vars)) throw new Error(`${templateFile}: no value for {{${name}}}`);
+    return vars[name];
+  });
+  return JSON.parse(filled);
+}
+
+/**
+ * GAP 1. `LocalSigner` and `SchnorrMultikey` both draw random `aux_rand` for a
+ * BIP340 signature, which a reproducible corpus cannot use. `Signer` is the
+ * published seam for exactly this: the update path depends on the interface and
+ * never on secret key bytes, so a signer that pins `aux_rand` to a label drops
+ * into `Updater.sign` with nothing else changed.
+ */
+class PinnedSigner implements Signer {
+  readonly publicKey: Uint8Array;
+  readonly #secretKey: Uint8Array;
+  readonly #auxRand: Uint8Array;
+
+  constructor(keyPair: SchnorrKeyPair, label: string) {
+    this.publicKey = keyPair.publicKey.compressed;
+    this.#secretKey = keyPair.secretKey.bytes;
+    this.#auxRand = sha256(utf8ToBytes(`did:btcr2 example corpus / aux_rand / ${label}`));
+  }
+
+  sign(data: Uint8Array, scheme: SigningScheme): Uint8Array {
+    if (scheme !== 'bip340') {
+      throw new Error(`PinnedSigner signs Data Integrity proofs only, not "${scheme}"`);
+    }
+    return schnorr.sign(data, this.#secretKey, this.#auxRand);
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Keys and Beacon endpoints
+// ---------------------------------------------------------------------------
+
+const keys = Object.fromEntries(
+  Object.entries(SEEDS).map(([name, seed]) => [name, seedKey(seed)]),
+) as Record<keyof typeof SEEDS, SchnorrKeyPair>;
+
+const network = getNetwork(NETWORK);
+
+/**
+ * GAP 4. `BeaconUtils.createBeaconServices` derives all three addresses of a
+ * key-based identifier from the one key in the identifier, which is what the
+ * key-based example needs and what it uses below. The external identifier gives
+ * each aggregate Beacon a separate aggregator key, so those endpoints are derived
+ * here and then handed back to `BeaconUtils.isBeaconService` to check.
+ */
+const endpoint = {
+  p2pkh: (pk: Uint8Array) => `bitcoin:${p2pkh(pk, network).address}`,
+  p2wpkh: (pk: Uint8Array) => `bitcoin:${p2wpkh(pk, network).address}`,
+  p2tr: (x: Uint8Array) => `bitcoin:${p2tr(x, undefined, network).address}`,
+};
+
+// ---------------------------------------------------------------------------
+// Genesis Document and the identifier derived from it
+// ---------------------------------------------------------------------------
+
+const genesisDocument = {
+  '@context': ['https://www.w3.org/ns/did/v1.1', 'https://btcr2.dev/context/v1'],
+  id: PLACEHOLDER,
+  verificationMethod: [
+    {
+      id: `${PLACEHOLDER}#key-0`,
+      type: 'Multikey',
+      controller: PLACEHOLDER,
+      publicKeyMultibase: keys.key0.publicKey.multibase.encoded,
+    },
+  ],
+  assertionMethod: [`${PLACEHOLDER}#key-0`],
+  capabilityInvocation: [`${PLACEHOLDER}#key-0`],
+  service: [
+    {
+      id: `${PLACEHOLDER}#service-0`,
+      type: 'SingletonBeacon',
+      serviceEndpoint: endpoint.p2wpkh(keys.key0.publicKey.compressed),
+    },
+    {
+      id: `${PLACEHOLDER}#service-1`,
+      type: 'CASBeacon',
+      serviceEndpoint: endpoint.p2tr(keys.casAggregator.publicKey.xOnly),
+    },
+    {
+      id: `${PLACEHOLDER}#service-2`,
+      type: 'SMTBeacon',
+      serviceEndpoint: endpoint.p2tr(keys.smtAggregator.publicKey.xOnly),
+    },
+  ],
+};
+
+for (const service of genesisDocument.service) {
+  if (!BeaconUtils.isBeaconService(service as BeaconService)) {
+    throw new Error(`${service.id} is not a Beacon Service`);
+  }
+}
+
+/** The implementation validates the Genesis Document and reduces it to genesis bytes. */
+const genesis = GenesisDocument.fromJSON(structuredClone(genesisDocument));
+const genesisBytes = GenesisDocument.toGenesisBytes(genesis);
+agree('genesis bytes', { hash: canonicalHash(genesisDocument) }, { hash: hashToBase64Url(genesisBytes) });
+
+/** An `x` HRP identifier commits to the Genesis Document hash. */
+const did = api.did.encode(genesisBytes, {
+  idType: IdentifierTypes.EXTERNAL,
+  version: BTCR2_VERSION_NUMBER,
+  network: NETWORK,
+});
+
+/** Resolution replaces the placeholder to establish version 1 of the document. */
+const initialDocument: Json = JSON.parse(
+  JSON.stringify(genesisDocument).replaceAll(PLACEHOLDER, did),
+);
+agree('version 1 document', genesis.toDidDocument(did), initialDocument);
+DidDocument.isValid(initialDocument);
+
+/** GAP 3 (property order). The specification lists `invocationTarget` before `controller`. */
+const derivedCapability = Appendix.deriveRootCapability(did);
+const rootCapability = {
+  '@context': derivedCapability['@context'],
+  id: derivedCapability.id,
+  invocationTarget: derivedCapability.invocationTarget,
+  controller: derivedCapability.controller,
+};
+agree('root capability', derivedCapability, rootCapability);
+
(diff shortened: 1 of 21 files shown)


src/data-structures.md
```diff
@@ -94,11 +94,16 @@ A [BTCR2 Unsigned Update] is a Map data structure with the following properties:
...
+  - `"https://w3id.org/json-ld-patch/v1"`
   - `"https://w3id.org/zcap/v1"`
   - `"https://w3id.org/security/data-integrity/v2"`
-  - `"https://w3id.org/json-ld-patch/v1"`
   - `"https://btcr2.dev/context/v1"`
+
+  Both the membership and the order of this array are byte-significant: changing either changes the
+  hash produced by the [JSON Document Hashing] algorithm, and that hash is what a
+  [BTCR2 Update Announcement] commits to on chain. The did-btcr2 method context is last on purpose
+  to ensure that the terms in `"https://btcr2.dev/context/v1"` with any conflicts between urls.
```
parasyte, line 106: Ah, yeah... JSC doesn't help with this. All arrays need to be sorted, including `@context`.

This applies to hashing the Genesis Document: `verificationMethod`, `assertionMethod`, `service`, all of them... The DID Document is technically the same document when the order of these arrays change, but hashing is order sensitive. And that's our constraint.

src/operations/resolve.md
```diff
@@ -202,9 +202,19 @@ Increment `current_version_id`.
...
+The resolver must find the entry of `current_document.capabilityInvocation` that identifies `update.proof.verificationMethod`. A reference entry identifies it when the two are equal; an embedded verification method object identifies it when the object's `id` is equal. Raise an [`INVALID_DID_UPDATE`] error if no entry does.
+
+Read `publicKeyMultibase` from that entry. When the entry is an embedded verification method object, read it from the object. When the entry is a reference, locate the verification method in `current_document.verificationMethod` whose `id` is equal to the reference and read `publicKeyMultibase` from there, raising an [`INVALID_DID_UPDATE`] error if there is no such verification method.
+
+Resolvers MUST NOT evaluate the proof's `created` or `expires` values. Resolution is a pure function of the DID, the resolution options, and the [Beacon Signal] and [Sidecar Data] inputs; wall-clock time is not one of those inputs. An implementation built on a generic Data Integrity {{#cite VC-DATA-INTEGRITY}} library MUST disable that library's expiry check.
```
parasyte, line 217: Is this always true? The Bitcoin blockchain pins the signed update to a block timestamp that probably should be evaluated against the proof. (The block timestamp is only a hint, but the network does some rudimentary sanity checks on it. ["Block times are accurate only to within an hour or two."](https://en.bitcoin.it/wiki/Block_timestamp) Should be more than enough accuracy for the proof validity window.)

I don't know if there are use cases where a signed update is expected to be withheld from publishing outside of the proof validity window, but that's the only legitimate situation I can imagine that would fail this kind of validation against the blockchain. In an adversarial scenario, the proof validity window may prevent some kinds of replay attacks.

src/data-structures.md
```diff
@@ -38,6 +38,10 @@ It can optionally include one or more of the following properties:
   - [BTCR2 Beacons][BTCR2 Beacon] are declared as DID services using the service `type` specified in
     [Beacons Table 1: Beacon Types].
 
+Verification method references in this document MAY be relative DID URLs, as permitted by DID Core
+v1.1 {{#cite DID-CORE}}. Implementations MUST resolve a relative DID URL against the document `id`
+before comparing it to another reference.
```
parasyte, line 43: I think this renegs on a compromise for a pain-point we had with the Rust implementation. Handling typed URLs properly requires handling internationalized domain names (IDNA), which in turn requires a zoo of Unicode tables and bloats the executable by several megabytes for something which should never be used by the resolver (to my knowledge).

"Treat URLs as strings" was the compromise we came up with. But handling relative URLs requires great care that "treat as a string" doesn't account for.

parasyte (commented):
I haven't read through the gen-examples script yet. But here are some initial thoughts!

src/operations/resolve.md
jintekc: I appreciate the push back here. I think disabling this feature is a cheap, easy but wrong path forward. I came to this conclusion after making the edits, pushing and opening the PR, so I figured I would wait and bring it up after a review.

I think the features of `created` and `expires` fall into the same category as the `versionTime` in #323 

> * `resolve.md:153` compares "the tuple's block time" without saying which timestamp. Bitcoin blocks carry two [timestamps]: the raw header timestamp (not monotonic, minable up to ~2h ahead) and median-time-past (MTP, monotonic non-decreasing, used by Bitcoin consensus for `locktime`). On a typical block, the two differ by tens of minutes, so resolvers comparing different fields select different versions near a boundary.
> [...]
> The comparison basis [should be] the `mediantime` (median-time-past) of the block containing the Beacon Signal transaction, as defined by Bitcoin Core.

Basically, I was internally inconsistent  between #323 and #324. The logic is #323 is: "MTP is an imperfect clock, but it is the same imperfect clock for every resolver; determinism over accuracy, in that sense." I think we should apply the SAME logic here.

One note to make. MTP is systematically behind the block's header timestamp because it's the median of the block and its 10 ancestors. That affects `expires` and `created` differently:

  - `expires`: MTP is the lenient direction. An honest update never falsely expire; it just gets ~an hour of slack making it safer.
  - `created`: MTP is the fatal direction. A proof created ten minutes before it was mined has created > `mediantime`, so a naive "created must not be after the block" check rejects nearly every honest update.

I think it makes most sense to evaluate `expires` against MTP and `created` against block header time (known as `nTime` in the RPC/API calls, I think).

I will re-work this language to do so. Thanks for the feedback @parasyte

src/data-structures.md
jintekc: Trying to get a better sense of the issue here so please bare with me.

If a document shows up with `"id": "#initialKey"` and `"capabilityInvocation": ["#initialKey"]`, what should a resolver do with it? That's the case I was trying to answer, and I don't think it goes away if we stay silent or forbid it. The document is valid per DID Core either way, so we're really choosing between resolvers normalizing it and resolvers rejecting it. For an x1 DID that rejection is permanent, since the genesis document is committed by hash.

Would it help if the only relative form allowed were a bare fragment? Then the rule is `document.id` plus `fragment`, no RFC 3986 machinery, and treating them as strings still holds. I wrote it broadly and I suspect that's the part actually causing grief, not `#initialKey` itself.

@danpape would like to get your thoughts on this given that it sounds like this would meaningfully impact the rust impl.

src/operations/resolve.md
parasyte: I thought the lock time (MTP?) was intended for contracts. I could easily be mistaken! Bitcoin is a very complex system. Resources claiming that lock time is used to prevent a broadcasted transaction from being included until a specific block is not really what I had in mind for pinning an update's "validity window" to the blockchain.

src/data-structures.md
parasyte: `#initialKey` is a nice and tidy relative URL, but it isn't representative of _all_ relative URLs allowed within a DID: https://www.w3.org/TR/did-1.1/#relative-did-urls

> Relative [DID URLs](https://www.w3.org/TR/did-1.1/#dfn-did-urls) _MAY_ contain relative path components, query parameters, and fragment identifiers.

This brings in a rather heavy URL handling requirement that isn't "just do string comparisons". And while that's OK to rely on - the Rust resolver will just have to suck it up and bring in the `url` crate - the side-effect is that the `url` crate brings in all of IDNA and its Unicode tables. This much is "a Rust problem".

But what I would *really* like to see somewhere is a decision record where we confirm whether or not IDNA support is necessary in the resolver. If the requirement is to support relative URLs, then *I think* IDNA can be disabled without causing interoperability problems. But I want to be sure. And I want it clearly determined whether resolvers are allowed to do that, or not.

danpape: I looked into it a bit, but we should ask Markus to confirm. DID 1.1 changed serviceEndpoint from "RFC 3986 URI" to "valid URL per the WHATWG URL Standard," and service ids and alsoKnownAs can be URL Standard URLs too. The URL Standard's host parser says to use UTS-46 ToASCII as part of parsing (and that spec is all about IDNAs) so I think any time we validate a non-DID URL in a document, we gotta deal IDNA. 

I do recall the "Treat URLs as strings" discussion Jay and I had one day, and I believe the argument was that maybe we don't need to validate the URLs, but now it seems like we do.

src/operations/resolve.md
jintekc: Locktime and MTP are two different things. MTP is a value that Bitcoin Core computes for every block and reports as `mediantime`. Locktime is a rule that reads it. Use of MTP here is use of that value only. It adds no dependency on locktime.

`expires` uses MTP because it is the security-relevant bound. It limits the time between the signature and the replay of a signed update. A miner sets the header timestamp and can increase it. No single miner can change MTP.

`created` uses the header timestamp because MTP is approximately one hour earlier than the header timestamp on mainnet. A controller signs the proof a few minutes before the block that contains it. A comparison of `created` against MTP therefore rejects almost every valid update. `created` is a sanity check and not a security boundary, because the controller selects the value and signs it.

Both values are available on the block. Neither adds a dependency.

src/data-structures.md
jintekc: I also looked into it a little, and here is what I came up with.

Q1. **Where does the IDNA requirement come from?**

From the validation of service endpoints, which @danpape identified. [DID Core, Services](https://www.w3.org/TR/did-1.1/#services) states:

> Each string value _MUST_ be a valid URL conforming to [URL Standard].

[URL Standard, host parser](https://url.spec.whatwg.org/#concept-host-parser) step 6 calls the domain parser:

> Let asciiDomain be the result of running domain parser with domain and false.

[URL Standard, IDNA](https://url.spec.whatwg.org/#domain-parser-toascii) then calls Unicode ToASCII:

> The domain parser ToASCII algorithm, given a scalar value string domain and a boolean beStrict, returns the result of running Unicode ToASCII with domain_name set to domain [...]

Two conditions are therefore necessary. The URL must have a host, and something must validate the URL.

---

Q2. **Does a did:btcr2 resolver have to do that validation?**

Only for beacon endpoints, and those have no host.

[Process Beacon Signals](https://github.com/dcdpr/did-btcr2/blob/main/src/operations/resolve.md?plain=1#L115) is the only place a resolver parses a `serviceEndpoint`:

> Scan the `service` entries in `current_document` [...] and identify BTCR2 Beacons by matching service `type` [...] Parse each beacon `serviceEndpoint` as a Beacon Address [...]

A [Beacon Address](https://github.com/dcdpr/did-btcr2/blob/main/src/terminology.md?plain=1#L37) is the Bitcoin address of a BTCR2 Beacon, which has no host. Every other `service` and `alsoKnownAs` value rides along in the document and never gets dereferenced.

@danpape I agree, good question for @peacekeeper, because DID Core section 6.1 is genuinely ambiguous about whether a consumer has to validate values it never dereferences. I think the specific question is: does the consumer duty in DID Core section 6.1 extend to full URL Standard validation of each `serviceEndpoint` value, or is it limited to the consumption of the representation?

---

Q3. **Do relative DID URLs require IDNA?**

It appears not. [DID Core, Relative DID URLs](https://www.w3.org/TR/did-1.1/#relative-did-urls) sets the base URI:

> When resolving a relative DID URL reference, the algorithm specified in RFC3986 Section 5: Reference Resolution _MUST_ be used. The base URI value is the DID that is associated with the DID subject.

A DID has no authority component, so the base contributes no host. The same section limits what the reference itself can contribute:

> Relative DID URLs _MAY_ contain relative path components, query parameters, and fragment identifiers.

A path, a query, and a fragment are the only parts in that list. None of them is a host. Neither side of the resolution can produce one, so the host parser never runs. This holds for relative path components, and not only for fragments.

---

Q4. **May a resolver disable IDNA?**

Yes. We can answer this in the spec by defining which checks a resolver does to meet the requirement that a DID document "MUST be conformant to DID Core v1.1".

Proposal to add to the spec:
- A did:btcr2 resolver does not need IDNA.
- A resolver MAY treat non-beacon `serviceEndpoint` and `alsoKnownAs` values as opaque strings.

peacekeeper:
Hi it would take me a bit more time to fully understand and comment on all the topics in this thread, but for now two quick thoughts:
- Somehow I don't think that in practice every resolver implementation out there does in-depth validation of WHATWG URL Standard conformance. To me it also feels like this is something that the RDF/JSON-LD layer should be doing, rather than making URI/URL handling a DID-specific topic.
- With regard to relative URI resolution, I do think that "string comparison" is not enough. If let's say you have `#initialKey` in `capabilityInvocation`, then you would have to resolve that against the base URI which is the DID itself, at least I think that's what we're doing in our Java implementation.

src/data-structures.md
danpape: > A [Beacon Address](https://github.com/dcdpr/did-btcr2/blob/main/src/terminology.md?plain=1#L37) is the Bitcoin address of a BTCR2 Beacon, which has no host

ooo, nice! it didn't quite dawn on me that this would only be an issue for the hostname part of the url, which it should have given the "DN" part of IDNA.

danpape (dismissed):

src/data-structures.md
```diff
@@ -94,11 +94,15 @@ A [BTCR2 Unsigned Update] is a Map data structure with the following properties:
...
+- `@context`: A context array. It MUST contain exactly the following context URLs, in this order:
+  - `"https://w3id.org/json-ld-patch/v1"`
   - `"https://w3id.org/zcap/v1"`
   - `"https://w3id.org/security/data-integrity/v2"`
-  - `"https://w3id.org/json-ld-patch/v1"`
   - `"https://btcr2.dev/context/v1"`
+
+  A change to the membership or the order of this array changes the hash that the
+  [JSON Document Hashing] algorithm produces. The did-btcr2 method context is last. Therefore its
+  terms replace the terms of the other contexts when two contexts define the same term.
```
parasyte: > Therefore its terms replace the terms of the other contexts when two contexts define the same term.

What does this mean? What's a term? Why are contexts being compared?

src/operations/resolve.md
```diff
@@ -202,9 +202,24 @@ Increment `current_version_id`.
...
+
+If `update.proof.created` or `update.proof.expires` is present, check each value against the Bitcoin block that contains the [Beacon Signal] that announced `update`. [^3] Raise an [`INVALID_DID_UPDATE`] error if either of the following conditions are not true:
+
+* `update.proof.created` is at or before the timestamp in the block header.
+* `update.proof.expires` is at or after the block `mediantime` {{#cite Bitcoin-Core}}.
+
+[^3]: In Data Integrity {{#cite VC-DATA-INTEGRITY}}, each method selects the time of interest for `created` and `expires`. On mainnet, the timestamp in the block header is approximately one hour later than `mediantime`. A controller signs a proof a short time before the block that contains it, so a check of `created` against `mediantime` rejects valid updates. For this reason, `created` uses the timestamp in the block header. A miner sets the timestamp in the header of its own block and can increase that value, but a single miner cannot change `mediantime`. The `expires` value limits the time between the signature and a replay of the update, so `expires` uses `mediantime`.
```
parasyte: > On mainnet, the timestamp in the block header is approximately one hour later than `mediantime`.

If `timestamp > mediantime`, then requiring `update.proof.created <= timestamp` is good, and `update.proof.expires >= mediantime` is also good. But there is still a corner case where `update.proof.expires < update.proof.created` is accepted. The expiration time is allowed to be approximately one hour earlier than the creation time.

Is this a specification gap, or is this requirement written somewhere that I missed?

src/data-structures.md
```diff
@@ -38,6 +38,18 @@ It can optionally include one or more of the following properties:
   - [BTCR2 Beacons][BTCR2 Beacon] are declared as DID services using the service `type` specified in
     [Beacons Table 1: Beacon Types].
 
+Verification method references in this document MAY be relative DID URLs. DID Core v1.1
+{{#cite DID-CORE}} permits this. Implementations MUST resolve a relative DID URL against the
+document `id` before they compare that URL with another reference. [^1]
+
+A resolver parses only one URL in a DID document. That URL is the `serviceEndpoint` of a
+[BTCR2 Beacon] service, and the resolver parses it as a [Beacon Address]. A resolver MAY treat all
+other URLs in the document as opaque strings. [^2]
```
parasyte: Style nitpick: Remove hard line breaks to match the rest of this commit?

bin/gen-examples.ts
```diff
@@ -0,0 +1,620 @@
...
+ * Nothing in the corpus is hand-written, and every step the reference
+ * implementation covers is performed by the reference implementation. The
+ * @did-btcr2/api SDK is the entry point and does as much as it reaches:
+ * identifiers, key pairs, multikeys, Data Integrity proofs, and the DID document
+ * classes it re-exports. The packages under it supply what it does not surface,
+ * listed at the import block. Every proof is verified and every hash is recomputed
+ * before anything is written, so a run that succeeds has verified the corpus
+ * rather than just serialized it. Where the implementation has not caught up to
+ * the specification, the specification wins and the difference is marked GAP.
+ * There are four.
```
parasyte: FWIW: I counted six "GAP"s in the script. "GAP 3" appears three times with different descriptions.

bin/gen-examples.ts
```diff
@@ -0,0 +1,620 @@
...
+ *
+ * SMT Proof Verification is the authority here, and `BTCR2MerkleTree` implements
+ * it: `bitAt(i)` is bit `i` of a 256-bit value counted from the least significant
+ * bit, so `i = 255` is the leaf level of the walk and `i = 0` is the root level.
+ * Two things follow. The algorithm does not state that reading, and `index` and
+ * `collapsed` are byte arrays by the time it uses them, so the text is worth
+ * pinning. And the worked example in Appendix: Optimized Sparse Merkle Tree
+ * Implementation counts the other way, which makes its `1101` proof unverifiable
+ * under this algorithm; the appendix is colloquial, so the example is what needs
+ * correcting.
```
parasyte, line 478: True enough about the Appendix. That chapter/section should probably be explicitly marked non-normative. It clearly uses pseudo-code but does not qualify it except for a brief note at the end.

parasyte (commented):
I think there is an ambiguity with the new update proof created/expires checks. The rest of my comments are less interesting.

src/data-structures.md
danpape: Yeah, this is just regular old JSON-LD processing rules and details which I don't think we need to mention here in this "list of properties" a BTCR2 Unsigned Update contains. I suggest we just remove lines 103-105.

jintekc: Yeah that's fair. its an over clarification to deepen the reasoning behind why exact ordering matters, but it probably adds more confusion than clarity.

src/operations/resolve.md
jintekc: This is a gap and a good catch. I can add a bullet to cover this corner case.

bin/gen-examples.ts
jintekc: Yeah I agree. Quite honestly, that section has caused a lot of issues. I will plan clarification changes to the Appendix SMT section in a follow up PR.

bin/gen-examples.ts
jintekc: Yeah, fair enough. All three instances of "GAP 3" have the same theme: a place where the TypeScript implementation produces JSON properties in a different order than the specification lists them in the Data Structures section. This is due to the TypeScript implementation being off from the spec. Upon re-read, I understand the confusion. I will see if I can make it more clear.

danpape (approved):

parasyte (approved):

PR #353, opened 2026-08-25
Replace the update private key input with a signer

### LINKED ISSUE #327: Remove raw private keys from the update operation
The spec shouldn't model passing raw keys through application code. The key's only use is being handed to the BIP340 cryptosuite ([update.md:89](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/update.md?plain=1#L89)), whose spec already accepts "a signing interface enabling the use of the private key bytes" in place of the key. Proposed edits:

1. Function signature ([update.md:16-24](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/update.md?plain=1#L16-L24)): replace `privateKey` with `signer`:

```rust
fn update(
  didSourceDocument,
  jsonPatches,
  targetVersionId,
  verificationMethodId,
  signer,
) ->
  signedUpdate
```

2. Inputs ([update.md:32-33](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/update.md?plain=1#L32-L33)): `signer` is a signing capability that returns a Schnorr signature over the given bytes using the private key associated with `verificationMethodId`. How it accesses that key is out of scope: an external signer is RECOMMENDED; holding the key in process is also conformant. No raw private key is an input.

3. Process overview ([update.md:44](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/update.md?plain=1#L44)): "signed with a private key" becomes "signed by the `signer`".

4. Construct BTCR2 Signed Update ([update.md:89](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/update.md?plain=1#L89)): instantiate the cryptosuite with the `"bip340-jcs-2025"` cryptosuite and `signer` as the signing interface. ("Signing interface" is the cryptosuite spec's own term; no change needed there.)

5. After the `createProof` step ([update.md:116](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/update.md?plain=1#L116)): add "Implementations SHOULD verify `update.proof` against the verification method's public key before announcing." A raw key lets you catch a key/verification-method mismatch before signing; an opaque signer doesn't, and announcing a mis-signed update permanently invalidates the DID (fatal `INVALID_DID_UPDATE`).

6. Output ([update.md:37](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/update.md?plain=1#L37)): the operation returns the signed update, which has no effect until announced; link "Announce DID Update" with an explicit anchor. The current wording oversells construction as anchoring.

7. If #326 is adopted, `deactivate()` takes `signer` the same way.

jintekc:
The private key's only use was the BIP340 Cryptosuite instance, and that cryptosuite already accepts a signing interface in place of the key bytes. `update` now takes `signer`, and the signer selects how it holds or reaches the key. An external signer is RECOMMENDED; an in-process key stays conformant.

Added a SHOULD for proof verification before announcement : a raw key let an implementation catch a mismatch between the key and the verification method. An opaque signer leaves that check to the implementation, and an announced update with an invalid proof permanently invalidates the DID.

The output states what the operation returns: a signed update that takes effect on resolution after its announcement.

Fixes #327

### DIFF
--- a/src/operations/update.md
+++ b/src/operations/update.md
@@ -18,7 +18,7 @@ fn update(
   jsonPatch,
   targetVersionId,
   verificationMethodId,
-  privateKey,
+  signer,
 ) ->
   signedUpdate
 ```
@@ -29,8 +29,7 @@ Input arguments:
 - `jsonPatch`: A single JSON Patch document {{#cite RFC6902}} with the changes to be made to the source DID document. Its wire shape is defined by the `patch` property of the [BTCR2 Unsigned Update (data structure)].
 - `targetVersionId`: The `versionId` that will be returned in the [DID document metadata (data structure)] once the new [BTCR2 Signed Update] is applied.
 - `verificationMethodId`: The `verificationMethod` ID used for signing the [BTCR2 Update].
-- `privateKey`: Private key associated with the `verificationMethodId`.
-  - An implementation MAY use the `verificationMethodId` ID to retrieve the private key from a key material manager.
+- `signer`: A signing interface. The signer receives bytes and returns a Schnorr signature {{#cite BIP340}} for those bytes. The signer makes the signature with the private key for `verificationMethodId`. The signer selects how it holds or reaches that key. An external signer is RECOMMENDED. An implementation that holds the key in its own process is also conformant.
 
 Outputs:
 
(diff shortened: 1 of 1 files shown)


danpape:
I like this in general. I understand what a "signing interface" could be, but both our spec and the BIP340 Cryptosuite spec are pretty terse explaining what that is or what it might look like. Should we add a few more details or just leave it as-is?

parasyte:
It's meant to be opaque. The only responsibility of a signer is to take some message and produce its signature.

A more complicated question is how a resolver acquires a signer in practice. The spec can safely delegate that complexity as out of scope.

jintekc:
> It's meant to be opaque. The only responsibility of a signer is to take some message and produce its signature.
> 
> A more complicated question is how a resolver acquires a signer in practice. The spec can safely delegate that complexity as out of scope.

Agreed

danpape (approved):

parasyte (approved):

Issue #262, opened 2025-09-20
Rename Intermediate Document to Template Document?


danpape:
For a while now I have felt uncomfortable about the name "Intermediate Document"--mainly because it seems like it is an ancestor of an "Initial Document".

Bryan created a mermaid diagram showing the "states" that a btcr2:did document could be in: https://github.com/dcdpr/website-btcr2-dev/blob/34a9a3736a4c4543f7dcc056526320136585f1df/docs/public/diagrams/did-document-state.mmd ... I'll copy it here for reference:

```mermaid
stateDiagram-v2
    state "Initial Document" as Initial
    state "Intermediate Document" as Intermediate
    state "Target Document" as Target
    state "Contemporary Document" as Contemporary
    state "Contemporary Document" as Contemporary1

    [*] --> Initial: Create From<br>Deterministic Key Pair
    [*] --> Intermediate: Create From<br>External DID Document
    Intermediate --> Initial: Replace placeholder DID
    Initial --> Contemporary: Resolve
    Contemporary --> Contemporary1: Traverse<br>Bitcoin Blockchain
    Contemporary1 --> Contemporary
    Contemporary1 --> Target
    Target --> [*]
```

Looking at it this way, it seems more clear to me that the name "Intermediate" is confusing. Given that that the point of the "Intermediate Document" is to provide a template containing DID placeholders to be replaced with real DIDs, I would like to propose that we rename "Intermediate Document" to "Template Document". Thoughts?

jintekc:
@parasyte great point. moving this issue from "close" to "keep"

PR #354, opened 2026-08-28
Add function signatures for `create` and `deactivate`

### LINKED ISSUE #326: Add function signatures for `create` and `deactivate`
Two of the four DID-method operations carry a function signature and two do not:

`resolve` has one ([resolve.md:15-18](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/resolve.md?plain=1#L15-L18)):

```rust
fn resolve(did, resolutionOptions) ->
  (didResolutionMetadata, didDocument, didDocumentMetadata)
```

`update` has one ([update.md:16-24](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/update.md?plain=1#L16-L24)):

```rust
fn update(
  didSourceDocument,
  jsonPatches,
  targetVersionId,
  verificationMethodId,
  privateKey,
) ->
  signedUpdate
```

`create` ([operations/create.md](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/create.md)) and `deactivate` ([operations/deactivate.md](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/deactivate.md)) have prose only.

This leaves the top-level operation surface half-specified: independent implementations' `create`/`deactivate` inputs and return shapes align only by chance. The spec should give all four operations a uniform Inputs/Outputs signature.

### PROPOSALS

1. `create` - encodes network + genesis bytes + version via the Identifier Encoding algorithm. Genesis bytes come from either a secp256k1 public key or a Genesis Document ([create.md:24-37](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/create.md?plain=1#L24-L37)):

```rust
fn create(
  genesisBytes,   // a secp256k1 public key or a Genesis Document hash
  network,
  version,
) ->
  did
```

For the key path the Initial DID Document is derived deterministically on first resolution; for the Genesis Document path the supplied document (with the `did:btcr2:_` placeholder id) is the genesis document distributed via sidecar/CAS.

2. `deactivate` - a specialized `update` whose patch is fixed to `{deactivated: true}`. This mirrors `update()` minus `jsonPatches`:

```rust
fn deactivate(
  didSourceDocument,
  targetVersionId,
  verificationMethodId,
  privateKey,
) ->
  signedUpdate
```

Keep `targetVersionId` caller-supplied, consistent with `update()`. Align `version`/`network` parameter typing with the Identifier Encoding algorithm's inputs (`version_number`, `network_name`; [algorithms.md:15-21](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/algorithms.md?plain=1#L15-L21)).

jintekc:
Two of the four DID-method operations carried a function signature and two did not. All four operations now use the same Inputs/Outputs form.

`create` takes the three values the identifier encodes. Each argument names the DID-BTCR2 Identifier Encoding input it supplies, so the two stay aligned.

`deactivate` is `update` with a fixed patch, so it takes the update arguments without `jsonPatch` and returns a signed update. Its signing input is `signer`, following #327.

Fixes #326

### DIFF
--- a/src/operations/deactivate.md
+++ b/src/operations/deactivate.md
@@ -5,4 +5,44 @@
 
 # Deactivate
 
-To deactivate a **did:btcr2**, the DID controller MUST add the property `deactivated` with the value `true` to the DID document. To do this, the DID controller constructs a valid [BTCR2 Update] with a JSON Patch {{#cite RFC6902}} that adds this property and announces the [BTCR2 Update] by broadcasting an [Authorized Beacon Signal] following the algorithms defined in [Update](update.md). Once a **did:btcr2** has been deactivated this state is considered permanent and resolution MUST terminate.
+To deactivate a **did:btcr2**, the DID controller MUST add the property `deactivated` with the value `true` to the DID document. Once a **did:btcr2** has been deactivated this state is considered permanent and resolution MUST terminate.
+
+The deactivate operation has the following function signature:
+
+```rust
+fn deactivate(
+  didSourceDocument,
+  targetVersionId,
+  verificationMethodId,
+  signer,
+) ->
+  signedUpdate
+```
+
+Input arguments:
+
+- `didSourceDocument`: The DID document being deactivated.
+- `targetVersionId`: The `versionId` that will be returned in the [DID document metadata (data structure)] once the new [BTCR2 Signed Update] is applied.
+- `verificationMethodId`: The `verificationMethod` ID used for signing the [BTCR2 Update].
+- `signer`: A signing interface, as defined for the [Update](update.md) operation.
+
+Outputs:
+
+- `signedUpdate`: A [BTCR2 Signed Update] with a patch that adds the `deactivated` property with the value `true`.
+
+
+## Process
+
+The deactivate operation is the [Update](update.md) operation with a predetermined patch:
+
+```json
+[
+  {
+    "op": "add",
+    "path": "/deactivated",
+    "value": true
+  }
+]
+```
+
+The DID controller constructs a valid [BTCR2 Update] with a JSON Patch {{#cite RFC6902}} that adds the `deactivated` property with the value `true`, and announces the [BTCR2 Update] by broadcasting an [Authorized Beacon Signal] following the algorithms defined in [Update](update.md).
--- a/src/operations/deactivate.md
+++ b/src/operations/deactivate.md
@@ -5,7 +5,7 @@
 
 # Deactivate
 
-To deactivate a **did:btcr2**, the DID controller MUST add the property `deactivated` with the value `true` to the DID document. Once a **did:btcr2** has been deactivated this state is considered permanent and resolution MUST terminate.
+To deactivate a **did:btcr2** identifier, the DID controller MUST add the property `deactivated` with the value `true` to the DID document. Once a **did:btcr2** identifier has been deactivated, this state is considered permanent and resolution MUST terminate.
 
 The deactivate operation has the following function signature:
 
--- a/src/appendix/privacy-considerations.md
+++ b/src/appendix/privacy-considerations.md
@@ -13,7 +13,7 @@
 
 ### DID Documents Can Be Private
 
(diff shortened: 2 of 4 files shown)


src/operations/deactivate.md
```diff
@@ -5,4 +5,32 @@
...
+- `signer`: A signing interface, as defined for the [Update](update.md) operation.
+
+Outputs:
+
+- `signedUpdate`: A [BTCR2 Signed Update] with a patch that adds the `deactivated` property with the value `true`.
+
+
+## Process
+
+The deactivate operation is the [Update](update.md) operation with a fixed patch. The DID controller constructs a valid [BTCR2 Update] with a JSON Patch {{#cite RFC6902}} that adds the `deactivated` property with the value `true`, and announces the [BTCR2 Update] by broadcasting an [Authorized Beacon Signal] following the algorithms defined in [Update](update.md).
```
parasyte: nit: I can't explain why, but using the word "fixed" to mean something like "statically defined" or "preselected" has always felt weird to me. Possibly because it's a homograph meaning "repaired".

parasyte (dismissed):

src/operations/deactivate.md
danpape: I thought I already made this comment here, but I guess I must have only said it out loud yesterday to @jintekc. I wanted to suggest that if we are going to say that the `deactivate` operation is just an `update` operation with a fixed (or statically defined) JSON patch, then we should probably show an example of what that fixed patch is rather than just mention it.

jintekc: @parasyte how about "predetermined" instead of "fixed" ?

@danpape I like the idea of showing the predetermined path as an inline JSON object. I will add this to my next push.

src/operations/deactivate.md
```diff
@@ -5,4 +5,32 @@
 
 # Deactivate
 
-To deactivate a **did:btcr2**, the DID controller MUST add the property `deactivated` with the value `true` to the DID document. To do this, the DID controller constructs a valid [BTCR2 Update] with a JSON Patch {{#cite RFC6902}} that adds this property and announces the [BTCR2 Update] by broadcasting an [Authorized Beacon Signal] following the algorithms defined in [Update](update.md). Once a **did:btcr2** has been deactivated this state is considered permanent and resolution MUST terminate.
+To deactivate a **did:btcr2**, the DID controller MUST add the property `deactivated` with the value `true` to the DID document. Once a **did:btcr2** has been deactivated this state is considered permanent and resolution MUST terminate.
```
jintekc: I think we are missing the word `identifier` a couple of times. See below.

> To deactivate a **did:btcr2** [identifier], the DID controller MUST add the property `deactivated` with the value `true` to the DID document. Once a **did:btcr2** [identifier] has been deactivated, this state is considered permanent and resolution MUST terminate.

src/operations/deactivate.md
```diff
@@ -5,4 +5,32 @@
...
+
+The deactivate operation has the following function signature:
+
+```rust
+fn deactivate(
+  didSourceDocument,
+  targetVersionId,
```
jintekc, line 15: a nit that has been bothering me for a while: can we rename `didSourceDocument` to `sourceDocument` or `sourceDidDocument`? Wedging the word `source` in between `did` and `Document` feels weird and somewhat wrong.

There is another 8 instances of the term `didSourceDocument` starting on line 17 in [update.md](https://github.com/dcdpr/did-btcr2/blob/b775bf83041716706e83382188e292526ce02cd4/src/operations/update.md?plain=1#L17).

I'd like to propose we swap all of those instances for either  `sourceDocument` or `sourceDidDocument`.

danpape: I wouldn't mind that. If you want to hold off on and add that in a more extensive renaming PR, there is also issue #262 that I've been meaning to get to. Or you can just knock it out here, of course.

src/operations/deactivate.md
danpape: This makes sense, looking at just this one paragraph, but maybe should check other uses of **did:btcr2** in other chapters to see if we are missing `identifier` elsewhere?

src/operations/deactivate.md
jintekc: fair enough. it can wait for the renaming PR. I'll add it as a comment in that issue.

src/operations/deactivate.md
jintekc: Should [privacy-considerations.md:16](https://github.com/dcdpr/did-btcr2/blob/b775bf83041716706e83382188e292526ce02cd4/src/appendix/privacy-considerations.md?plain=1#L16) have one too?

> [...] To keep the DID document fully private, the DID controller can choose to use an externally resolved initial **did:btcr2** [identifier] and not place the initial DID document on a [Content Addressable Storage] (CAS) system such as the InterPlanetary File System (IPFS). [...]

danpape (approved):

parasyte (approved):

Issue #342, opened 2026-08-11
Define resolve outcomes for aggregate Beacon Signals


jintekc:
## Problem

Process SMT Beacon ([resolve.md:146](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/operations/resolve.md?plain=1#L146)) uses `smt_proof.updateId` as `update_hash` unconditionally. However, `updateId` is OPTIONAL ([data-structures.md:255](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/data-structures.md?plain=1#L255)). SMT Proof Verification has an explicit branch for its absence ([algorithms.md:169-173](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/algorithms.md?plain=1#L169-L173)). Therefore a verified no-update proof reaches [resolve.md:131-135](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/operations/resolve.md?plain=1#L131-L135) with `update_hash` unset. The resolver then raises `MISSING_UPDATE_DATA`. Thus a controller that publishes no update in a signal causes a fatal error.

Three more cases have no defined outcome:

* an `smt_root` with no entry in `smt_lookup_table`;
* a proof whose verification returns false;
* in Process CAS Beacon ([resolve.md:141](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/operations/resolve.md?plain=1#L141)), an announcement with no entry for the `did`. [beacons.md:55](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/beacons.md?plain=1#L55) says that this absence expresses CAS non-inclusion.

## Proposed Solution

Every Beacon Signal that the scan finds resolves to exactly one of four outcomes:

* Verified proof, `updateId` present: set `update_hash` and continue.
* Verified proof, `updateId` absent, or a verified announcement with no entry for the `did`: the signal announces nothing for this DID. Skip the signal. Append no tuple, and raise no error.
* Evidence missing (neither a proof nor an announcement is available, or an `smt_root` has no entry in `smt_lookup_table`): raise `MISSING_UPDATE_DATA`.
* Evidence present, but it fails verification against the signal: raise `INVALID_SIGNAL_DATA` (new code, supporting edit 2).

The resolver skips a signal only on verified evidence. If a missing proof means "nothing in this signal for me", then a controller can conceal any update when it withholds one proof. That is the Late Publishing that this design prevents.

[...]

PR #357, opened 2026-09-01
Require hash verification on every CAS retrieval

### LINKED ISSUE #333: Require hash verification on every CAS retrieval
## Problem

During resolution, there are three places where content can be retrieved from CAS. Only one of them requires additional verification of the retrieved bytes: the genesis document. Once retrieved using the hash decoded from the x1 DID, the bytes are hashed and checked to match, otherwise raise an `INVALID_DID` error if the hash does not match `genesis_bytes` ([resolve.md:68](https://github.com/dcdpr/did-btcr2/blob/079cdd9/src/operations/resolve.md?plain=1#L68)). The update retrieval ([resolve.md:134](https://github.com/dcdpr/did-btcr2/blob/079cdd9/src/operations/resolve.md?plain=1#L134)) and the announcement retrieval ([resolve.md:141](https://github.com/dcdpr/did-btcr2/blob/079cdd9/src/operations/resolve.md?plain=1#L141))  do not verify. A gateway can substitute the content, and the resolver does not detect the substitution. Downstream checks catch most substituted updates. Downstream checks do not always catch a substituted CAS Announcement.

## Proposed Solution

Add one requirement to [update-data-distribution.md](https://github.com/dcdpr/did-btcr2/blob/079cdd9/src/update-data-distribution.md): for each retrieval from CAS, the resolver MUST compute the SHA-256 hash of the retrieved content. The resolver MUST compare the computed hash to the hash that was used for the retrieval. If the hashes do not match, the resolver MUST NOT use the content. The retrieval then fails.

Each path keeps its own error for a failed retrieval:
- Genesis retrieval: `NOT_FOUND` or `INVALID_DID`, unchanged.
- Update retrieval: `MISSING_UPDATE_DATA`; the update "is not available from either source" ([resolve.md:135](https://github.com/dcdpr/did-btcr2/blob/079cdd9/src/operations/resolve.md?plain=1#L135)).
- Announcement retrieval: the outcomes are defined in #342.

Each retrieval site points back to the requirement added to update data distribution by stating "as described in [BTCR2 Update Data Distribution]" to the retrieve-from-CAS sentence at [resolve.md:68](https://github.com/dcdpr/did-btcr2/blob/079cdd9/src/operations/resolve.md?plain=1#L68), [resolve.md:134](https://github.com/dcdpr/did-btcr2/blob/079cdd9/src/operations/resolve.md?plain=1#L134), and [resolve.md:141](https://github.com/dcdpr/did-btcr2/blob/079cdd9/src/operations/resolve.md?plain=1#L141). The general retrieval paragraph ([resolve.md:70](https://github.com/dcdpr/did-btcr2/blob/079cdd9/src/operations/resolve.md?plain=1#L70)) already points there.

jintekc:
Three places in the resolve operation retrieve content from CAS, and only one of them checks the retrieved bytes against the hash used to retrieve them. A gateway could substitute content without the resolver noticing.

The requirement to check the retrieved bytes against the hash used to retrieve them now lives once in BTCR2 Update Data Distribution. Three CAS retrieval references in the spec now point at that one requirement.

Added a link reference for `[BTCR2 Update Data Distribution]` since it wasn't previously defined. The one existing reference renders as a literal bracket text in the built book.

Fixes #333

### DIFF
--- a/src/operations/resolve.md
+++ b/src/operations/resolve.md
@@ -65,7 +65,7 @@ raised while decoding.
 - Hash each [CAS Announcement (data structure)] in `sidecar.casUpdates` with the [JSON Document Hashing] algorithm and build a map from hash to announcement (`cas_lookup_table`).
 - Build a map from `sidecar.smtProofs` keyed by proof `id` (`smt_lookup_table`).
 
-If `genesis_bytes` is a SHA-256 hash, hash `sidecar.genesisDocument` with the [JSON Document Hashing] algorithm. If `sidecar.genesisDocument` is not provided, retrieve it from [CAS] using `genesis_bytes`. Raise a [`NOT_FOUND`] error if the [Genesis Document] cannot be retrieved. Raise an [`INVALID_DID`] error if the computed hash does not match `genesis_bytes`.
+If `genesis_bytes` is a SHA-256 hash, hash `sidecar.genesisDocument` with the [JSON Document Hashing] algorithm. If `sidecar.genesisDocument` is not provided, retrieve it from [CAS] using `genesis_bytes` as described in [BTCR2 Update Data Distribution]. Raise a [`NOT_FOUND`] error if the [Genesis Document] cannot be retrieved. Raise an [`INVALID_DID`] error if the computed hash does not match `genesis_bytes`.
 
 When data is not available in [Sidecar Data], implementations are RECOMMENDED to retrieve it from a [Content Addressable Storage][CAS] ([CAS]) service. To retrieve a document from [CAS], construct a CID from the document's SHA-256 hash bytes as described in [BTCR2 Update Data Distribution].
 
@@ -131,14 +131,14 @@ For each transaction found:
 * Build a tuple with:
   * The transaction's block metadata (height, time, and confirmations).
   * The [BTCR2 Signed Update (data structure)] retrieved from `update_lookup_table[update_hash]`.
-    * If the update is not in `update_lookup_table`, retrieve it from [CAS].
+    * If the update is not in `update_lookup_table`, retrieve it from [CAS] using `update_hash` as described in [BTCR2 Update Data Distribution].
     * Raise a [`MISSING_UPDATE_DATA`] error if the update is not available from either source.
 * Append the tuple to `updates`.
 
 
 ### Process CAS Beacon { #process-cas-beacon }
 
-Treat [Signal Bytes] as `map_update_hash`. Look up `map_update_hash` in `cas_lookup_table` to retrieve a [CAS Announcement (data structure)] and read `update_hash` from the announcement entry keyed by `did`. If the [CAS Announcement (data structure)] is not in `cas_lookup_table`, retrieve it from [CAS].
+Treat [Signal Bytes] as `map_update_hash`. Look up `map_update_hash` in `cas_lookup_table` to retrieve a [CAS Announcement (data structure)] and read `update_hash` from the announcement entry keyed by `did`. If the [CAS Announcement (data structure)] is not in `cas_lookup_table`, retrieve it from [CAS] using `map_update_hash` as described in [BTCR2 Update Data Distribution].
 
 
 ### Process SMT Beacon { #process-smt-beacon }
--- a/src/update-data-distribution.md
+++ b/src/update-data-distribution.md
@@ -47,6 +47,8 @@ on a network for retrieval based on its content, not its name or location. The c
 address is determined by a cryptographic hash of the file. The hash is then passed into 
 a retrieval function specific to the type of [CAS] to retrieve the file.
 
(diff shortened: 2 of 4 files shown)


src/operations/resolve.md
```diff
@@ -131,14 +131,14 @@ For each transaction found:
 * Build a tuple with:
   * The transaction's block metadata (height, time, and confirmations).
   * The [BTCR2 Signed Update (data structure)] retrieved from `update_lookup_table[update_hash]`.
-    * If the update is not in `update_lookup_table`, retrieve it from [CAS].
+    * If the update is not in `update_lookup_table`, retrieve it from [CAS] as described in [BTCR2 Update Data Distribution].
```
danpape: line 68 above says "retrieve it from [CAS] using `genesis_bytes`" but here we just say "retrieve it from [CAS]" ... maybe we say "retrieve it from [CAS] using `update_hash`" ?

src/operations/resolve.md
```diff
@@ -131,14 +131,14 @@ For each transaction found:
...
     * Raise a [`MISSING_UPDATE_DATA`] error if the update is not available from either source.
 * Append the tuple to `updates`.
 
 
 ### Process CAS Beacon { #process-cas-beacon }
 
-Treat [Signal Bytes] as `map_update_hash`. Look up `map_update_hash` in `cas_lookup_table` to retrieve a [CAS Announcement (data structure)] and read `update_hash` from the announcement entry keyed by `did`. If the [CAS Announcement (data structure)] is not in `cas_lookup_table`, retrieve it from [CAS].
+Treat [Signal Bytes] as `map_update_hash`. Look up `map_update_hash` in `cas_lookup_table` to retrieve a [CAS Announcement (data structure)] and read `update_hash` from the announcement entry keyed by `did`. If the [CAS Announcement (data structure)] is not in `cas_lookup_table`, retrieve it from [CAS] as described in [BTCR2 Update Data Distribution].
```
danpape: line 68 above says "retrieve it from [CAS] using genesis_bytes" but here we just say "retrieve it from [CAS]" ... maybe we say "retrieve it from [CAS] using update_hash" ?

danpape: We probably also need a "Raise a XXX error ..."

src/update-data-distribution.md
```diff
@@ -47,6 +47,8 @@ on a network for retrieval based on its content, not its name or location. The c
 address is determined by a cryptographic hash of the file. The hash is then passed into 
 a retrieval function specific to the type of [CAS] to retrieve the file.
 
+For each retrieval from [CAS], the resolver MUST compute the SHA-256 hash of the retrieved content. The resolver MUST compare the computed hash to the hash that was used for the retrieval. If the hashes do not match, the resolver MUST NOT use the content. The retrieval then fails.
```
danpape: Hmm, here we say "The retrieval then fails." ... but make no mention of an error to raise. But maybe that is good, because then it leaves it up to the calling site o decide which error ought to be raised.

danpape (changes requested):

src/operations/resolve.md
jintekc: yeah that's a good call. create consistency across each instance.

src/operations/resolve.md
jintekc: assuming you mean "retrieve it from [CAS] using `map_update_hash`" ? Will also add this.

jintekc: yeah you're right. we should have another sentence saying "Raise a [`MISSING_UPDATE_DATA`] error if the update is not available from either source."

jintekc: actually, I almost forgot I specifically left out the error here. See rationale in #333 and #342.

src/update-data-distribution.md
jintekc: > then it leaves it up to the calling site o decide which error ought to be raised.

Exactly. That's the rationale for not specifying an error.

* genesis bytes section "Raise a [NOT_FOUND](https://www.w3.org/TR/did-resolution-1.0/#errors) error if the [Genesis Document](http://localhost:3000/terminology.html#genesis-document) cannot be retrieved. Raise an [INVALID_DID](https://www.w3.org/TR/did-resolution-1.0/#errors) error if the computed hash does not match genesis_bytes."
* update section "Raise a [MISSING_UPDATE_DATA](http://localhost:3000/errors.html#missing_update_data) error if the update is not available from either source."
* cas beacon is missing the error per my above rationale about #333 and #342 

Because this part of the spec deals with a boundary between our spec and an external CAS system, we shouldn't specify any one error about that failure since the request to CAS could fail for many reasons. 

Honestly, I think the simplest answer is to remove "The retrieval then fails." Its fluff that adds nothing.

danpape: I get it now. Sounds good.

peacekeeper (dismissed):
As discussed in the thread, I agree we should consistently say "using `genesis_bytes`", "using `update_hash`", etc... But otherwise looks good.

danpape (approved):

parasyte (approved):

PR #358, opened 2026-09-02
Masthead document metadata

### LINKED ISSUE #197: Add More Metadata to Spec Masthead

danpape:
I added a "This Document" block to the masthead so the spec carries its own published URL, its source repository, its commit history, and where to file feedback.

The block follows the DID Resolution spec's masthead as its model.

Fixes #197

### DIFF
--- a/src/index.md
+++ b/src/index.md
@@ -36,7 +36,7 @@ WARNING This specification is still under active development and may be subject 
 
 ## Publication Date
 
-... 2025
+... 2026
 
 ## License
 
@@ -44,4 +44,4 @@ WARNING This specification is still under active development and may be subject 
 
 ## Copyright
 
-&copy; 2024-2025 Digital Contract Design
+&copy; 2024-2026 Digital Contract Design
--- a/src/index.md
+++ b/src/index.md
@@ -36,7 +36,7 @@ WARNING This specification is still under active development and may be subject 
 
 ## Publication Date
 
-... 2026
+{{ publication_date }}
 
 ## License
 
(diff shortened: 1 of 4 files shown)


src/index.md
```diff
@@ -24,14 +24,24 @@ WARNING This specification is still under active development and may be subject
...
+| This version   | [https://dcdpr.github.io/did-btcr2/](https://dcdpr.github.io/did-btcr2/)                           |
+| Repository     | [https://github.com/dcdpr/did-btcr2](https://github.com/dcdpr/did-btcr2)                           |
+| Commit history | [https://github.com/dcdpr/did-btcr2/commits/main](https://github.com/dcdpr/did-btcr2/commits/main) |
+| Feedback       | [https://github.com/dcdpr/did-btcr2/issues](https://github.com/dcdpr/did-btcr2/issues)             |
+
+
 ## Publication Date
 
-... 2025
+... 2026
```
jintekc, line 40: What does "Publication Date" mean exactly? I read it as "the date it was first published", and if that's true, then should we set this to be a specific, unchaning date based on when we first published this latest rewrite version?

danpape: I don't recall the initial decision, but I gather maybe we left that as a placeholder for the day we "release" 1.0?

danpape: ... the w3c specs stamp the date they were last re-generated from source. Maybe we can do something like that.

jintekc: Yeah I think that would be a good idea.  I think that would mean every time we merge a PR into main and re-publish.

danpape: ok, I added some tera stuff that will insert the date when the book is rendered, but we won't really see it until we merge this PR to main.

jintekc (dismissed):

jintekc (approved):

parasyte (approved):

PR #359, opened 2026-09-04
Define the versionId and versionTime resolution semantics

### LINKED ISSUE #323: Clarify versionId/versionTime resolution semantics

jintekc:
Fixes #323.

`versionId` and `versionTime` are mutually exclusive per DID Resolution v1. A request with both raises `INVALID_OPTIONS`. The issue proposed a new `INVALID_RESOLUTION_OPTIONS` error. The DID Resolution definition of `INVALID_OPTIONS` covers this case, so the spec reuses that code instead. An unsatisfiable `versionId` raises `NOT_FOUND`. The `versionTime` comparison uses the block `mediantime` with an inclusive boundary, the same basis as the `expires` check. The `versionId` stop test now runs before Apply, so version 1 is reachable.

Two supporting edits come with the step reorder. The `block_confirmations` step now runs after the `versionTime` stop, so the metadata reflects a tuple that the resolver processed. The resolve chapter footnotes now follow document order.

The shipped `resolution-options.json` example carried both selectors. The generator now emits `versionId` only, and I regenerated the corpus.

### DIFF
--- a/src/operations/resolve.md
+++ b/src/operations/resolve.md
@@ -26,6 +26,8 @@ Raise an [`INVALID_OPTIONS`] error if `resolutionOptions` contains both `version
 
 [^1]: DID Resolution v1 {{#cite DID-RESOLUTION}} defines the two options as mutually exclusive.
 
+When provided, `resolutionOptions.versionId` MUST be parsed as an integer and `resolutionOptions.versionTime` SHOULD be parsed as an XML Datetime. Raise an [`INVALID_OPTIONS`] error if either value does not parse.
+
 Resolution maintains the following state while building the DID document:
 
 * `updates`: a list of tuples, each containing Bitcoin block metadata (height, mediantime, confirmations) and a [BTCR2 Signed Update (data structure)].
@@ -152,7 +154,7 @@ Treat [Signal Bytes] as `smt_root`. Look up `smt_root` in `smt_lookup_table` to 
 
 ## Process `updates` Array { #process-updates }
 
-1. If `resolutionOptions.versionId` is provided and `current_version_id` is equal to the integer form of `resolutionOptions.versionId`, resolve `current_document` as `didDocument`.
+1. If `resolutionOptions.versionId` is provided and `current_version_id` is equal to the parsed `resolutionOptions.versionId`, resolve `current_document` as `didDocument`.
 2. If `updates` is empty or `current_document.deactivated` is `true`:
     * Raise a [`NOT_FOUND`] error if `resolutionOptions.versionId` is provided.
     * Otherwise, resolve `current_document` as `didDocument`.
--- a/src/operations/resolve.md
+++ b/src/operations/resolve.md
@@ -22,9 +22,13 @@ fn resolve(did, resolutionOptions) ->
 
 Input values MUST first go through [Decoding the DID](#decode-the-did) and [Processing Sidecar Data](#process-sidecar-data).
 
+Raise an [`INVALID_OPTIONS`] error if `resolutionOptions` contains both `versionId` and `versionTime`. [^1]
+
+[^1]: DID Resolution v1 {{#cite DID-RESOLUTION}} defines the two options as mutually exclusive.
+
 Resolution maintains the following state while building the DID document:
 
-* `updates`: a list of tuples, each containing Bitcoin block metadata (height, time, confirmations) and a [BTCR2 Signed Update (data structure)].
+* `updates`: a list of tuples, each containing Bitcoin block metadata (height, mediantime, confirmations) and a [BTCR2 Signed Update (data structure)].
 * `current_document`: the DID document being assembled.
 * `current_version_id`: the version number being processed (starts at `1`).
 * `update_hash_history`: a list of [BTCR2 Unsigned Update] hashes used to detect duplicates.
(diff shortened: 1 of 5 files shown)


danpape (dismissed):

src/operations/resolve.md
```diff
@@ -148,16 +152,16 @@ Treat [Signal Bytes] as `smt_root`. Look up `smt_root` in `smt_lookup_table` to
 
 ## Process `updates` Array { #process-updates }
 
-Resolve `current_document` as `didDocument` if `updates` is empty.
-
-Otherwise:
+1. If `resolutionOptions.versionId` is provided and `current_version_id` is equal to the integer form of `resolutionOptions.versionId`, resolve `current_document` as `didDocument`.
```
parasyte: `resolutionOptions.versionId` is a string (and `resolutionOptions.versionTime` is, too). It's implied that comparisons are done against *parsed* versions of these strings, ~~but I don't see it explicitly called out after this patch. It used to say in the steps below:~~ (I'm clearly not reading what I'm quoting. This already says "integer form".)

> If `current_version_id` is greater than or equal to the integer form of `resolutionOptions.versionId`

With "integer form" doing most of the work. Lexicographical comparisons like `"10" > "2"` results in `false`! Integer parsing MUST be done on `versionId`.

For `versionTime`, we just happen to get lucky with ISO 8601 formatted UTC timestamps. Lexicographical comparisons on these do the right thing, at least until the year 10,000, when it breaks again. :joy:

I recommend making sure the spec is clear that `versionId` MUST be parsed as an integer for comparisons. For `versionTime`, string comparisons are OK. But a spec-compliant resolver SHOULD parse the timestamp, not just verify that it looks like a reasonable datetime.

Just my 2 cents.

jintekc: All good call outs. Agreed. Thanks for the 2 cents. I'll take it! :smile:

parasyte: I made the last paragraph in my old comment stronger: "`versionId` should be parsed" => "`versionId` MUST be parsed". To reflect what I actually meant to say.

jintekc: I added a sentence at the top under "Process" stating the versionId MUST be parsed as an integer and versionTime SHOULD be parsed as an XML Datetime. The below step references the "parsed `resolutionOptions.versionId`" because of the MUST, but I did not include "parsed" before `resolutionOptions.versionTime` since its a SHOULD, not a hard MUST requirement.

parasyte (approved):
LGTM

danpape (approved):

PR #363, opened 2026-09-06
Terminate the resolve loop and rename its subsections

### LINKED ISSUE #360: Terminate the resolve loop
## Problem

The resolver repeats one loop: "Process Beacon Signals", then "Process updates Array" ([resolve.md:36-39](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L36-L39)). "Process updates Array" handles one tuple per pass. It sorts `updates` and takes the first tuple ([resolve.md:155](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L155)). It does not remove that tuple from `updates`. "Process Beacon Signals" scans every beacon of `current_document` on every pass. It appends every signal that it finds to `updates` ([resolve.md:117](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L117), [resolve.md:136](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L136)). No rule skips a signal that an earlier pass appended.

As a result, the loop takes the same tuple on every pass. Example: one beacon announces the updates for versions 2, 3, and 4. Pass 1 takes the version 2 tuple and applies it. Pass 2 appends the three tuples again, sorts, and takes the version 2 tuple, now a duplicate. Every later pass does the same. The loop never reaches version 3, and `updates` never becomes empty. For a plain `resolve(did, {})` on an active DID, the empty `updates` rule ([resolve.md:151](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L151)) is the only exit, so a literal implementation does not terminate. A resolver that clears `updates` before each scan has the same problem, because the sort selects the version 2 tuple again.

## Proposed Solution

Make each Beacon Signal enter `updates` exactly once, and remove each tuple that the loop takes. The number of signals is finite, so `updates` becomes empty after a finite number of passes. The existing empty `updates` rule then resolves the document.

Add one item to the resolution state ([resolve.md:27-31](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L27-L31)): `scanned_beacons`, a list of the Beacon Addresses that "Process Beacon Signals" already scanned (starts empty).

In "Process Beacon Signals" ([resolve.md:117](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L117)), skip a Beacon Address that is in `scanned_beacons`. Scan the Beacon Addresses that are not in the list, then add each of them to `scanned_beacons`. One scan of an address finds every signal of that address in the block chain, so a second scan finds nothing new. A later pass scans only the addresses that an applied update added to the document. That is the reason for the one-tuple-per-pass shape: an update can add or remove beacons, so the scan must run again after each update.

In "Process updates Array" ([resolve.md:155](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L155)), change "Take the first tuple." to "Take the first tuple and remove it from `updates`."

## Supporting edits

1. **Reword the loop bullets.** At [resolve.md:37-38](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L37-L38), change "populate `updates`" to "add tuples to `updates`", and change "apply updates to `current_document`" to "apply one update to `current_document`". The current words suggest that one pass applies every update.

2. **Delete the cache hint.** At [resolve.md:119](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L119), delete "Cache Beacon Addresses to avoid repeated transaction lookups." The skip rule in the Solution replaces it.

### LINKED ISSUE #362: Rename "process updates array"
Our current resolve description says:

The resolver:

1. Establishes current_document from the DID or from Sidecar Data.
2. Repeats the following loop:
  * Process Beacon Signals to populate updates from the beacon services in current_document.
  * Process updates Array to apply updates to current_document and refresh block_confirmations.
  * The loop terminates when processing updates resolves didDocument or an error occurs.

but the name (and description of later in the document) "Process updates Array" isn't very clear that really all it is doing is dealing with only the NEXT update that needs to be applied. 

I was thinking it might read better if we called it "Process next update"? I wouldn't mind renaming "Process Beacon Signals" to something like "Gather updates from Beacon Signals".

Unless anyone is opposed, I'll work up a PR for review.

jintekc:
The resolve loop never terminates as written. Process `updates` Array selects one tuple per pass and never removes it. Process Beacon Signals scans every beacon of `current_document` on every pass and appends every signal again. So the loop selects the same tuple on every pass and never reaches the next version. Removal of the tuple alone is not enough. The next scan appends the same signals again, and the sort puts the tuple of the applied version first on every pass.

The resolver now scans each Beacon Address once. A new `scanned_beacons` state list holds the scanned addresses. Process Beacon Signals finds transactions only for a Beacon Address that is not in the list, and adds the address to the list. Step 3 removes the first tuple from `updates`. The number of signals is finite, so `updates` becomes empty after a finite number of passes. The empty `updates` rule in step 2 then resolves the document. The one-tuple-per-pass shape stays, because an applied update can add or remove beacons.

Two supporting edits. The loop bullets now say that one pass adds tuples and applies one update. The "cache Beacon Addresses" hint is gone. It implied that the resolver scans an address once, and the `scanned_beacons` condition now states it.

The second commit renames the two loop subsections, as Dan suggested in the issue. Process Beacon Signals becomes Find Beacon Signals, and Process `updates` Array becomes Process Next Update. The anchors follow the headings, `#find-beacon-signals` and `#process-next-update`, so a link to an old anchor lands at the top of the page.

Fixes #360.
Fixes #362.

### DIFF
--- a/src/operations/resolve.md
+++ b/src/operations/resolve.md
@@ -31,6 +31,7 @@ When provided, `resolutionOptions.versionId` MUST be parsed as an integer and `r
 Resolution maintains the following state while building the DID document:
 
 * `updates`: a list of tuples, each containing Bitcoin block metadata (height, mediantime, confirmations) and a [BTCR2 Signed Update (data structure)].
+* `scanned_beacons`: a list of [Beacon Addresses][Beacon Address] that [Process Beacon Signals](#process-beacon-signals) scanned (starts empty).
 * `current_document`: the DID document being assembled.
 * `current_version_id`: the version number being processed (starts at `1`).
 * `update_hash_history`: a list of [BTCR2 Unsigned Update] hashes used to detect duplicates.
@@ -40,8 +41,8 @@ The resolver:
 
 1. [Establishes `current_document`](#establish-current-document) from the DID or from [Sidecar Data].
 2. Repeats the following loop:
-    * [Process Beacon Signals](#process-beacon-signals) to populate `updates` from the beacon services in `current_document`.
-    * [Process `updates` Array](#process-updates) to apply updates to `current_document` and refresh `block_confirmations`.
+    * [Process Beacon Signals](#process-beacon-signals) to add tuples to `updates` from the beacon services in `current_document`.
+    * [Process `updates` Array](#process-updates) to apply one update to `current_document` and refresh `block_confirmations`.
     * The loop terminates when processing updates resolves `didDocument` or an error occurs.
 
 The resolver returns:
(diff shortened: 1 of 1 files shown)


src/operations/resolve.md
```diff
@@ -118,11 +119,16 @@ Render the [Initial DID Document] template with these values (Bitcoin addresses
...
+Scan the `service` entries in `current_document` ([DID Document (data structure)]) and identify [BTCR2 Beacons][BTCR2 Beacon] by matching service `type` to [Beacons Table 1: Beacon Types]. Parse each beacon `serviceEndpoint` as a [Beacon Address].
 
-Implementations are RECOMMENDED to query an indexed Bitcoin blockchain Remote Procedure Call (RPC) service such as [electrs](https://github.com/romanz/electrs) or [Esplora](https://github.com/Blockstream/esplora). Implementations MAY instead traverse blocks from the genesis block. Cache [Beacon Addresses][Beacon Address] to avoid repeated transaction lookups.
+For each [Beacon Address] that is not in `scanned_beacons`:
+
+* Find the Bitcoin transactions that spend from the [Beacon Address] and whose last output script contains [Signal Bytes].
+* Add the [Beacon Address] to `scanned_beacons`.
```
parasyte, line 129: I see, this elevates the original address cache recommendation to a hard spec requirement. I'm OK with that.

parasyte (approved):

src/operations/resolve.md
jintekc: Exactly! 🤝

danpape (approved):

PR #364, opened 2026-09-06
Do not stop at `versionTime` for a duplicate update

### LINKED ISSUE #361: Skip the `versionTime` stop for a duplicate update
## Problem

"Process updates Array" sorts `updates` by `targetVersionId`, then by block height, and takes the first tuple ([resolve.md:155](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L155)). Step 3 then stops resolution if the tuple's block time is more recent than `resolutionOptions.versionTime` ([resolve.md:157](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L157)). The step runs before the resolver knows if the tuple is a duplicate of the current version. A duplicate of the current version in a later block sorts before the next version, because the sort key is `targetVersionId` first.

Example: the version 2 update is announced at block 100 and again at block 300. The version 3 update is announced at block 150. The request specifies a `versionTime` between the times of blocks 150 and 300. The sorted tuples are (version 2, block 100), (version 2, block 300), and (version 3, block 150). The resolver applies the version 2 update from block 100 and then takes the duplicate at block 300. Its block time is more recent than `versionTime`, so the resolver returns version 2. Version 3 was valid at `versionTime`, so version 3 is the correct result. The spec expects duplicate announcements: Confirm Duplicate Update ([resolve.md:175-181](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L175-L181)) exists for that case. The example assumes that each processed tuple leaves `updates`. See #360.

## Proposed Solution

Add one sentence at the start of step 3 ([resolve.md:157](https://github.com/dcdpr/did-btcr2/blob/5eeee40/src/operations/resolve.md?plain=1#L157)). The step then applies only to a tuple that advances the version. The existing sentence stays unchanged:

"Skip this step if the tuple's `targetVersionId` is not greater than `current_version_id`. If `resolutionOptions.versionTime` is provided and the tuple's block time is more recent, resolve `current_document` as `didDocument`."

A duplicate of the current version then skips the time test and reaches Confirm Duplicate Update. A true duplicate passes that check, and the loop continues with the next tuple. In the example, the loop reaches the version 3 tuple at block 150 and applies it. `updates` is then empty, and the resolver returns version 3.

One outcome changes. Today a duplicate after `versionTime` with the same `targetVersionId` and a different hash stops resolution silently, because the time test runs first. With the new sentence, that tuple reaches Confirm Duplicate Update and raises `LATE_PUBLISHING`. That is the correct outcome for an equivocal history. A resolution without `versionTime` raises the same error.

jintekc:
Process Next Update sorts `updates` by `targetVersionId` first and then by block height. The sort puts a duplicate of version 2 before version 3, and the block heights do not change this sequence. Step 4 stops resolution if the block `mediantime` of the tuple is after `versionTime`. The resolver does that test before it knows that the tuple is a duplicate. Thus a duplicate in a block after `versionTime` stops resolution, and the resolver does not apply version 3 from a block before `versionTime`. The resolver resolves an incorrect version.

With this change, step 4 stops resolution at `versionTime` only for a tuple with a `targetVersionId` that is more than `current_version_id`. A duplicate has a `targetVersionId` that is equal to `current_version_id`, so step 4 does not stop resolution for it. The duplicate continues to Confirm Duplicate Update, and the loop then continues to version 3. The resolver applies version 3 if its block is before `versionTime`. A duplicate with a different hash in a block after `versionTime` continues to Confirm Duplicate Update and raises `LATE_PUBLISHING`. A resolution without `versionTime` raises the same error.

Fixes #361.

### DIFF
--- a/src/operations/resolve.md
+++ b/src/operations/resolve.md
@@ -165,7 +165,10 @@ Treat [Signal Bytes] as `smt_root`. Look up `smt_root` in `smt_lookup_table` to 
     * Raise a [`NOT_FOUND`] error if `resolutionOptions.versionId` is provided.
     * Otherwise, resolve `current_document` as `didDocument`.
 3. Sort `updates` by [BTCR2 Signed Update (data structure)] `targetVersionId` (ascending) with the tuple's block height as a tiebreaker. Remove the first tuple from `updates`.
-4. If `resolutionOptions.versionTime` is provided and the tuple's block `mediantime` {{#cite Bitcoin-Core}} is after `resolutionOptions.versionTime`, resolve `current_document` as `didDocument`. [^4]
+4. Resolve `current_document` as `didDocument` if all of the following conditions are true:
+    * The tuple's `targetVersionId` is more than `current_version_id`.
+    * `resolutionOptions.versionTime` is provided.
+    * The tuple's block `mediantime` {{#cite Bitcoin-Core}} is after `resolutionOptions.versionTime`. [^4]
 5. Set `block_confirmations` to the tuple's block confirmations.
 6. Set `update` to the tuple's [BTCR2 Signed Update (data structure)] and [check `update.targetVersionId`](#check-update-version).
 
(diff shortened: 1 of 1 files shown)


parasyte (dismissed):
I had to run the algorithm in my head to make sense of why this was correct. But it looks good.

This could do with a short comment explaining the purpose of the new condition. (I know we can always reference the blame and find this PR again, but that's an indirect way to justify the existence of what looks like an unrelated condition.) To wit, this is only needed because duplicate updates are explicitly allowed. The existence of duplicate updates makes this condition compulsory.

parasyte (approved):

danpape (approved):
Looks good!

PR #365, opened 2026-09-08
Specify the SMT leaf values, the proof bit sequence, and the aggregate signal results

### LINKED ISSUE #177: Ensure that all sidecarDocuments and smtProofs have been used if targetVersionX is not defined.
The absence of both `targetVersionId` and `targetVersionTime` means "get the latest version of the DID document". At the end of processing, both `sidecarDocuments` and `smtProofs` should be empty. This should be checked for.

https://github.com/dcdpr/did-btc1/blob/5a645c970545bf17ceca9ba7b9da61d9fe7dfd3b/chapters/Beacons.md?plain=1#L488-L499

### LINKED ISSUE #341: Define SMT leaf values for both `nonce` modes

### LINKED ISSUE #342: Define resolve outcomes for aggregate Beacon Signals

### LINKED ISSUE #343: Pin the byte-level details of SMT proof verification
## Problem

Two conformant implementations can compute different roots from the same proof. SMT Proof Verification and the SMT Proof data structure leave three things undefined:

* The bit order of `bitAt`. The algorithm calls `bitAt` twice ([algorithms.md:180](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/algorithms.md?plain=1#L180), [algorithms.md:186](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/algorithms.md?plain=1#L186)). Only the closed [PR #320](https://github.com/dcdpr/did-btcr2/pull/320) defines `bitAt`. The spec text has no definition.
* The width of `collapsed`, and the number of entries that `hashes` must contain.
* The meaning of the entries of `hashes`. The description at [data-structures.md:257](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/data-structures.md?plain=1#L257) lists values that a proof never carries.

A test against real bytes is also impossible, because every value in the example proof is a placeholder stub. The Solution below covers the first gap. The supporting edits cover the other gaps.

## Solution

Add the definition of `bitAt` to the algorithm text. Use the semantics that the [#320 review thread](https://github.com/dcdpr/did-btcr2/pull/320#discussion_r3431301269) already states: the bit index increases from left to right over the decoded bytes. Thus `bitAt(0)` is the most significant bit of the first byte, and `bitAt(255)` is the least significant bit of the last byte.

The definition must be in the spec because the other interpretation is equally possible. `index = int(hash(did))` ([optimized-smt.md:117](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/appendix/optimized-smt.md?plain=1#L117)) suggests a bit index that starts at the least significant bit. The two interpretations bit-reverse every leaf position. Thus two implementations compute different roots from the same proof.

Under the left-to-right definition, the verifier loop (`let i = 255 - n`, [algorithms.md:178](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/algorithms.md?plain=1#L178)) consumes the least significant bit at the leaf level. That agrees with the appendix walkthrough ([optimized-smt.md:224-225](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/appendix/optimized-smt.md?plain=1#L224-L225)). Under the least-significant-bit-first interpretation, the loop and the walkthrough disagree. One sentence or one pseudocode comment settles the question. The review thread suggested both forms. Neither form reached the spec.

## Supporting edits

1. **State the width of `collapsed` and its length invariant.** [data-structures.md:256](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/data-structures.md?plain=1#L256) says only "Bitmap of zero nodes within the path". The verifier loops exactly 256 times. Thus `collapsed` is 256 bits, and `len(hashes) + popcount(collapsed)` must equal 256. Define the error that the verifier raises when the invariant fails. Otherwise `proof.hashes` underflows with no defined behavior. Cite the definition in the spec itself ([optimized-smt.md:126](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/appendix/optimized-smt.md?plain=1#L126)) instead of the external project.

2. **Correct the `hashes` description.** [data-structures.md:257](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/data-structures.md?plain=1#L257) says that the entries contain the hash of the signed update or the zero identity. A proof carries only the hashes of the non-empty sibling nodes. The verifier recovers the zeros from `collapsed`. The leaf-to-root order that the description states is correct and stays.

3. **Replace the placeholder proof with a byte-valid example.** Every value in [sidecar-smt-proof.json:2-9](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/example-data/sidecar-smt-proof.json#L2-L9) is a `<< base64url ... >>` stub. The appendix at [optimized-smt.md:202-212](https://github.com/dcdpr/did-btcr2/blob/1f314af/src/appendix/optimized-smt.md?plain=1#L202-L212) repeats the same stub values. Generate the example after the spec defines the leaf value encodings. Include a no-nonce non-inclusion proof, because three of the four leaf forms have no example proof anywhere in `src/`. #102 tracks the fail-state vectors separately.

jintekc:
- **#341**

  The two arms of SMT Proof Verification hash `proof.nonce`, but `nonce` is OPTIONAL. A proof without a `nonce` did not verify. The fields `nonce` and `updateId` select one of four leaf values: `hash(hash(nonce) + updateId)`, `hash(hash(nonce))`, the bare `updateId`, or `cachedZero[0]` for an empty index. `hash()` accepts a byte sequence of any length, and `concat()` accepts two 32-byte values. A decoded `updateId` or entry of `hashes` that is not 32 bytes makes the result `false`. Without this rule, a zero-length `updateId` makes the first arm equal to the second. The `nonce` has no length rule, because `hash(nonce)` is 32 bytes for a `nonce` of any length. Find Beacon Signals now has a MUST for the resolver: hash the retrieved update and compare the hash to `update_hash`. The appendix, the SMT Proof data structure, and Aggregate Beacons now align with the four arms.

- **#343**

  The algorithm did not give the bit sequence of `bitAt(i)`. Two implementations can read the bits of `hash(did)` in a different sequence and calculate different roots from the same proof. `bitAt(0)` is the most significant bit of the first byte, and the walk starts at the leaf with `bitAt(255)`. `collapsed` is 32 bytes, and the number of entries in `hashes` plus the number of `1` bits in `collapsed` must be 256. The committed example proof used the other bit sequence, and this PR replaces the SMT examples. The examples have two proofs from a second SMT signal: an update with a `nonce`, and an empty index without one.

- **#342**

  The resolver had no result for a signal with no update for `did`, or with a proof that did not verify. A signal has four results: `update_hash`, no update for `did`, `MISSING_UPDATE_DATA`, or the new `INVALID_SIGNAL_DATA`. `MISSING_UPDATE_DATA` applies if the announcement or the proof is not available, and `INVALID_SIGNAL_DATA` applies if the proof does not verify. `smtProofs` holds one proof for each SMT signal that Find Beacon Signals finds. Find Beacon Signals finds only the signals at or above `current_block_height`, the block of the update that added the Beacon Address.

- **#177**

  The specification did not give the effect of Sidecar Data that the resolver did not use. This data has no effect on the result, if the resolution options have no `versionId` and no `versionTime`. It can show that the resolver and the DID controller do not read the same Bitcoin blocks, or a problem with the Sidecar Data. Implementations MAY tell the caller which Sidecar Data they did not use.

Fixes #341.
Fixes #343.
Fixes #342.
Fixes #177.


src/operations/resolve.md
```diff
@@ -171,7 +183,7 @@ Treat [Signal Bytes] as `smt_root`. Look up `smt_root` in `smt_lookup_table` to
     * The tuple's `targetVersionId` is more than `current_version_id`. [^4]
     * `resolutionOptions.versionTime` is provided.
     * The tuple's block `mediantime` {{#cite Bitcoin-Core}} is after `resolutionOptions.versionTime`. [^5]
-6. Set `block_confirmations` to the tuple's block confirmations.
+6. Set `block_confirmations` to the tuple's block confirmations. Set `current_block_height` to the tuple's block height.
```
danpape: This step 6 sets both variables for every tuple, but they are defined (in the "Resolution maintains the following state" section above) as the block of the most recently *applied* update, and whether the tuple applies is decided in step 7.

I'd suggest moving both variable assignments into Apply `update`, next to the `update_hash_history` append. Step 6 then goes away.

src/operations/resolve.md
danpape: We should either add " and `current_block_height`" or maybe just leave out the whole "and refresh..." part?

src/operations/resolve.md
jintekc: Good catch, agreed.  Step 6 sets both variables for a duplicate tuple, and Confirm Duplicate Update applies nothing. Then `confirmations` comes from the block of the duplicate, and footnote 2 specifies the lowest block height. We can remove step 6 and put the two sentences after the `update_hash_history` sentence in the "Apply update" section.

src/operations/resolve.md
jintekc: Yeah, I think we should leave out the whole "and refresh `block_confirmations`" part.

src/algorithms.md
```diff
@@ -142,16 +142,16 @@ Example output:
 
 To verify the inclusion or non-inclusion of a DID in the [SMT Proof], perform the following steps:
 
-Throughout this section, `hash()` denotes SHA-256 {{#cite SHA256}} over a byte array, and `concat()` (equivalently, the `+` operator) concatenates byte arrays, not strings. The `base64url` {{#cite RFC4648}} encoded fields of an [SMT Proof (data structure)] (`id`, `nonce`, `updateId`, `collapsed`, and the entries of `hashes`) MUST be decoded to their raw bytes before being used in any of these operations.
+Throughout this section, `hash()` denotes SHA-256 {{#cite SHA256}} over a byte sequence. The byte sequence has no length limit. `concat()` (equivalently, the `+` operator) concatenates two 32-byte values into one 64-byte value. `0` denotes 32 zero bytes. `bitAt(i)` of a 32-byte value counts from left to right. `bitAt(0)` is the most significant bit of the first byte. `bitAt(255)` is the least significant bit of the last byte. The `base64url` {{#cite RFC4648}} encoded fields of an [SMT Proof (data structure)] (`id`, `nonce`, `updateId`, `collapsed`, and the entries of `hashes`) MUST be decoded to their raw bytes before being used in any of these operations.
```
danpape: We are talking a lot about bytes now, but in later examples in this section and elsewhere, `hash(did)` takes a string, right? The DID is ASCII, so it can only mean one thing, but since this paragraph is now the place that specifies every byte-level detail, maybe we should add something like: "`hash(did)` is `hash()` over the UTF-8 bytes of the DID string."

src/data-structures.md
```diff
@@ -235,7 +235,7 @@ The [Sidecar Data] contains optional properties:
 - `casUpdates`: OPTIONAL array of [CAS Announcements][CAS Announcement (data structure)]. It is REQUIRED
   if the DID being resolved has used a [CAS Beacon] to publish a [BTCR2 Update].
 - `smtProofs`: OPTIONAL array of [SMT Proofs][SMT Proof (data structure)]. It is REQUIRED
-  if the DID being resolved has used a [SMT Beacon] to publish a [BTCR2 Update].
+  if any version of the DID document being resolved includes an [SMT Beacon]. It MUST contain one [SMT Proof (data structure)] for each [Beacon Signal] of an [SMT Beacon] that [Find Beacon Signals](operations/resolve.md#find-beacon-signals) finds.
```
danpape: How about a DID whose latest update added an SMT Beacon, but the cohort has not signalled yet. The beacon is in the document, so this clause requires `smtProofs`, but there is no proof to put in it, so we seem to be saying it is required to have an empty array here. The next sentence (one proof for each signal that Find Beacon Signals finds) implies this optional array be present when proofs are available, so I think the REQUIRED clause could go.

src/data-structures.md
```diff
@@ -262,19 +262,27 @@ An [SMT Proof] data structure contains the following properties:
...
 - `updateId`: The OPTIONAL [BTCR2 Signed Update (data structure)] hashed with the [JSON Document Hashing] algorithm.
-- `collapsed`: Bitmap of zero nodes within the path (see: [collapsed leaves](https://github.com/hoytech/quadrable#collapsed-leaves)). MUST be `"base64url"` {{#cite RFC4648}} encoded without padding.
-- `hashes`: Array of SHA-256 hashes representing the sibling [SMT] nodes from the leaf, containing the SHA-256 hash of the [BTCR2 Signed Update] or the "zero identity", to the root. Each hash MUST be `"base64url"` {{#cite RFC4648}} encoded without padding.
+- `collapsed`: 256-bit bitmap with one bit for each level of the path from the leaf to the root. A `1` bit identifies a level at which the sibling is an empty subtree. [Appendix: Optimized Sparse Merkle Tree Implementation] gives the value of an empty subtree. Bit `255` is the leaf level and bit `0` is the root level, in the bit sequence that [SMT Proof Verification] specifies. MUST be 32 bytes `"base64url"` {{#cite RFC4648}} encoded without padding. The number of entries in `hashes` plus the number of `1` bits in `collapsed` MUST be `256`.
```
danpape: Someones code could put `cachedZero[n]` in `hashes` and leave some or all of the `collapsed` bits clear, and the proof would still verify. To make sure they use `collapsed` correctly, should we add something like "A proof writer SHOULD set the `collapsed` bit rather than include the empty subtree's value in `hashes`."

src/data-structures.md
jintekc: If the cohort has not signaled, doesn't that imply that there is no Beacon Signal on-chain with minconf=6 yet for that signaling round? And there would be nothing found for that SMT beacon during the "Find Beacon Signal" operation?

danpape: Right, but before this PR's diff, it said "It is REQUIRED if the DID being resolved has used a [SMT Beacon] to publish a [BTCR2 Update]." and after the diff it says "It is REQUIRED if any version of the DID document being resolved includes an [SMT Beacon]." ... not "includes and published an update". 

So I think that for a window of time, we are saying that there is a requirement for storing an array that is empty. ... If we simply remove the "It is REQUIRED" sentence, then the description collapses to "`smtProofs`: OPTIONAL array of ... that MUST contain one [SMT Proof (data structure)] ..." Meaning, if there are no proofs to be found (for CAS reasons or early SMT reasons), then the array doesn't even need to exist since it is OPTIONAL, but if there are proofs around, then the array will exist and MUST have the proofs in it.   

small nit I guess, but I thought I'd point it out.

jintekc: Ah yeah, I see now. This change specifically addresses a DID in an SMT cohort that announces no update. That DID must still provide a proof for each signal, in this case its a proof of non-inclusion. Without that proof, Process SMT Beacon finds no entry in the `smt_lookup_table` and raises a `MISSING_UPDATE_DATA` error.

That said, you make a good point about there being a state in between adding the beacon and cohort not signaling yet. We can remove the "It is REQUIRED ..." sentence and it should resolve that.

> - `smtProofs`: OPTIONAL array of [SMT Proofs][SMT Proof (data structure)]. It MUST contain one [SMT Proof (data structure)] for each [Beacon Signal] of an [SMT Beacon] that [Find Beacon Signals](operations/resolve.md#find-beacon-signals) finds.

src/algorithms.md
jintekc: Yes, `did` is a string, and it is the only input that the algorithm hashes as a string. Maybe we can add your sentence after "The byte sequence has no length limit":

> `hash(did)` denotes `hash()` over the UTF-8 bytes of the DID string.

src/data-structures.md
jintekc: yeah, that's a good catch. Probably a good idea to upgrade the language for clarity. I think we can replace 

>  A `1` bit identifies a level at which the sibling is an empty subtree. 

with

 > A 1 bit identifies a level at which the sibling is an empty subtree. The bit SHOULD be 1 at each of those levels.
 
Thoughts?

danpape: Sounds good to me. I'd love to hear an opinion from @parasyte

parasyte: The collapsing is "just" a compression technique because those zero values are deterministic. Both forms are valid proofs. I don't have any preference for enforcing compression by SMT proof producers. But @rxgrant might.

----

Unrelated to this thread, the one thing that stands out about this change is that it doesn't define the endianness for "bit `0` and bit `255`". Unlike algorithms.md, which is pretty clear about it:

> `bitAt(i)` of a 32-byte value counts from left to right. `bitAt(0)` is the most significant bit of the first byte. `bitAt(255)` is the least significant bit of the last byte.

jintekc: > Unrelated to this thread, the one thing that stands out about this change is that it doesn't define the endianness for "bit 0 and bit 255". Unlike algorithms.md, which is pretty clear about it:
> 
> bitAt(i) of a 32-byte value counts from left to right. bitAt(0) is the most significant bit of the first byte. bitAt(255) is the least significant bit of the last byte.

Solid point. I agree, we should define endianness wherever necessary to ensure clarity.

src/data-structures.md
```diff
@@ -262,19 +261,27 @@ An [SMT Proof] data structure contains the following properties:
 SHA-256 hashes {{#cite SHA256}} (`id`, `updateId`, `hashes`) MUST be `"base64url"` {{#cite RFC4648}} encoded without padding.
 
 - `id`: SHA-256 hash of the root node.
-- `nonce`: OPTIONAL 256-bit nonce generated for each update. MUST be encoded as a string using `"base64url"` {{#cite RFC4648}} encoding without padding.
+- `nonce`: OPTIONAL 256-bit nonce, one for each index in each [Beacon Signal]. MUST be encoded as a string using `"base64url"` {{#cite RFC4648}} encoding without padding. The DID controller keeps each `nonce` for the life of the DID. If the DID controller does not have the `nonce`, the [SMT Proof] of that [Beacon Signal] cannot be verified.
```
danpape: Shouldn't the controller keep the whole proof, not just the `nonce`? From the `smtProofs` thread above, it has to hand the resolver a proof for every signal, including rounds where it announced nothing. It can keep its own `nonce`, but `id`, `collapsed` and `hashes` come from the aggregator's tree, and nothing says the aggregator keeps old trees around. Maybe this belongs on the `smtProofs` bullet instead, as something like "The DID controller keeps each [SMT Proof] for the life of the DID"?

jintekc: Yeah, that's a fair point. I guess I was thinking about earlier discussions around aggregation services keeping proofs on behalf of participants, which is part of the appeal of a service: shared cost and data retention. The nonce was left out of that for privacy. However, that did not made its way into the spec anywhere, and [aggregate-beacons.md:127](https://github.com/dcdpr/did-btcr2/blob/be67bce/src/beacons/aggregate-beacons.md?plain=1#L127) already has the participant persisting the proofs. I think this is a topic for discussion later though. Not sure where it belongs since it falls into the non-normative aggregate-beacons section.

That said, I'm good with your suggestion. How about we add this sentence to the `smtProofs` bullet:

> - `smtProofs`: OPTIONAL array of [SMT Proofs][SMT Proof (data structure)]. It MUST contain one [SMT Proof (data structure)] for each [Beacon Signal] of an [SMT Beacon] that [Find Beacon Signals](operations/resolve.md#find-beacon-signals) finds. The DID controller keeps each [SMT Proof (data structure)] for the life of the DID.

Then, remove a sentence from `nonce`:

> - `nonce`: OPTIONAL 256-bit nonce, one for each index in each [Beacon Signal]. MUST be encoded as a string using `"base64url"` {{#cite RFC4648}} encoding without padding. If the DID controller does not have the `nonce`, the [SMT Proof] of that [Beacon Signal] cannot be verified.

Keeping the last sentence there since only the controller ever has the nonce, so I think its important to call that out.

src/data-structures.md
jintekc: @parasyte upon further review, I realized that the bullet itself doesn't define the bit order, but it points to [SMT Proof Verification](https://github.com/dcdpr/did-btcr2/blob/4746417/src/algorithms.md?plain=1#L147), which does:

> `bitAt(i)` of a 32-byte value counts from left to right. `bitAt(0)` is the most significant bit of the first byte. `bitAt(255)` is the least significant bit of the last byte.

and the pseudocode reads `collapsed` with that same `bitAt(i)` ([algorithms.md:224](https://github.com/dcdpr/did-btcr2/blob/4746417/src/algorithms.md?plain=1#L224)). I'd rather keep one definition and point to it than restate it in data-structures.md. Two copies of the same sentence drift eventually, which is the ["keep redundancies to an absolute minimum"](https://github.com/dcdpr/did-btcr2/pull/345#discussion_r3777094570) point from #345.

The current sentence "in the bit sequence that [SMT Proof Verification] specifies" doesn't name `bitAt(i)`. An alternative is to make the pointer exact. How about:

> Bit `255` is the leaf level and bit `0` is the root level. Bit `i` is `bitAt(i)` of the decoded value as specified in the [SMT Proof Verification] algorithm.

However, I am not married to anything, so if you'd prefer we restate it here, I am okay with that too.

src/data-structures.md
danpape: Looks good to me, thanks

src/data-structures.md
parasyte: Referencing `bitAt(i)` seems fine to me.

danpape (approved):

danpape (approved):

src/data-structures.md
rxgrant: Allowing the option looks like a source of proof mutability under validating conditions.  We do not currently know how that could hurt us, but it should be disallowed just in case.

src/operations/resolve.md
```diff
@@ -140,23 +146,29 @@ For each transaction found:
...
+
+Read `update_hash` from the announcement entry keyed by `did`. If the announcement has no entry for `did`, the [Beacon Signal] announces no update for `did`.
 
 
 ### Process SMT Beacon { #process-smt-beacon }
 
-Treat [Signal Bytes] as `smt_root`. Look up `smt_root` in `smt_lookup_table` to retrieve an [SMT Proof (data structure)]. Validate the proof with the [SMT Proof Verification] algorithm. Use `smt_proof.updateId` as `update_hash`.
+Treat [Signal Bytes] as `smt_root`. Look up `smt_root` in `smt_lookup_table` to retrieve an [SMT Proof (data structure)] as `smt_proof`. Raise a [`MISSING_UPDATE_DATA`] error if `smt_lookup_table` has no entry for `smt_root`. Raise an [`INVALID_SIGNAL_DATA`] error if the `id` of `smt_proof` is not equal to `smt_root`.
+
+Verify `smt_proof` with the [SMT Proof Verification] algorithm. Raise an [`INVALID_SIGNAL_DATA`] error if the result of the algorithm is `false`. If `smt_proof` has an `updateId`, use it as `update_hash`. If `smt_proof` has no `updateId`, the [Beacon Signal] announces no update for `did`.
```
parasyte, line 171: From my [first comment](https://github.com/dcdpr/did-btcr2/pull/365#issuecomment-5623985631):

> Additionally, `updateId` is required to match the Signed Update hash, but I don't actually see where that is specified (forgive me if this PR already fixes that). It is implied by the normative SMT Proof data structure. This requirement forces `updateId` to be a valid hash, so it should already be a 32-byte array.

The aggregate-beacons chapter used to say `json_document_hash(update)`, which is as close as it ever got to making the connection, AFAICT.

parasyte (changes requested):

src/operations/resolve.md
parasyte: Ah, I see it hiding on line 156! So, this was already addressed, just not where I expected it to be. I think it makes sense where it is, since it also covers CAS Beacons.

parasyte (approved):

PR #367, opened 2026-09-10
Ignore the signals of a removed Beacon Address

### LINKED ISSUE #366: Ignore the Beacon Signals of a Beacon Address after an update removes the address
## Problem

[Beacons](https://github.com/dcdpr/did-btcr2/blob/8591478ae6686da61ca6025eebd5877d2eaa6f9a/src/beacons.md?plain=1#L10) specifies that all Beacon Signals from a BTCR2 Beacon in the Current DID Document "MUST be processed". This requirement does not include a signal from a Beacon Address that is not in the Current DID Document. A BTCR2 Update can remove a Beacon Address from the DID document. The resolver applies that update mid-resolution, but [Find Beacon Signals](https://github.com/dcdpr/did-btcr2/blob/8591478ae6686da61ca6025eebd5877d2eaa6f9a/src/operations/resolve.md?plain=1#L128) still found the Beacon Signals of that removed address. No step removes the tuples of the removed address from `updates` mid-resolution. Thus the resolver will apply an update from a Beacon Signal that is no longer in the current DID document.

Example:

1. Version 1 of the DID document has the Beacon Address `A`.
2. Block 100: a Beacon Signal of `A` announces version 2.
3. Block 200: a Beacon Signal of `A` announces version 3. Version 3 removes `A` from the DID document and adds the Beacon Address `B`.
4. Block 300: a Beacon Signal of `A` announces version 4.

Find Beacon Signals examines `A` one time and finds the signals in blocks 100, 200, and 300. The resolver applies version 2, version 3, and version 4. But at block 300, `A` is not in the Current DID Document. The signal in block 300 is not a signal of a BTCR2 Beacon in the Current DID Document. Thus the resolver must not apply version 4 from that signal.

## Solution

Remove the tuples of an Beacon Address that is also removed. This change keeps the version-first sort of Process Next Update and adds one step.

1. Add the Beacon Address of the transaction to each tuple that Find Beacon Signals makes.
2. After "Apply update" changes `current_document`, compare the Beacon Addresses of `current_document` before and after the patch. For each Beacon Address that the patch removes:
    * Remove from `updates` each tuple with that Beacon Address and a block height more than `current_block_height`.
    * Remove the Beacon Address from `scanned_beacons`. A new update can add the address again.

jintekc:
Find Beacon Signals scans each Beacon Address one time and finds all of its signals, until the last block. If an update removes that address from the DID document, the signals of the address in the blocks after the removal stay in `updates`. The resolver applies them. Those signals are not Authorized Beacon Signals, because the address is not in the DID document at that time. Thus the resolver can resolve an incorrect version.

With this change, each tuple in `updates` includes the Beacon Address of its signal. A new step 4 of Process Next Update ignores the tuple if `current_document` has no BTCR2 Beacon with that address. The step runs before the `versionTime` step, so an ignored tuple does not stop resolution and does not set `block_confirmations`. The resolver applies only the signals of the Beacon Addresses in the current DID document.

Fixes #366.

### DIFF
--- a/src/operations/resolve.md
+++ b/src/operations/resolve.md
@@ -30,7 +30,7 @@ When provided, `resolutionOptions.versionId` MUST be parsed as an integer and `r
 
 Resolution maintains the following state while building the DID document:
 
-* `updates`: a list of tuples, each containing Bitcoin block metadata (height, mediantime, confirmations) and a [BTCR2 Signed Update (data structure)].
+* `updates`: a list of tuples, each containing Bitcoin block metadata (height, mediantime, confirmations), a [Beacon Address], and a [BTCR2 Signed Update (data structure)].
 * `scanned_beacons`: a list of [Beacon Addresses][Beacon Address] that [Find Beacon Signals](#find-beacon-signals) scanned (starts empty).
 * `current_document`: the DID document being assembled.
 * `current_version_id`: the version number being processed (starts at `1`).
@@ -142,6 +142,7 @@ For each transaction found:
   * [SMT Beacon]: use [Process SMT Beacon](#process-smt-beacon).
 * Build a tuple with:
   * The transaction's block metadata (height, mediantime, and confirmations).
+  * The [Beacon Address] of the transaction.
   * The [BTCR2 Signed Update (data structure)] retrieved from `update_lookup_table[update_hash]`.
     * If the update is not in `update_lookup_table`, retrieve it from [CAS] using `update_hash` as described in [BTCR2 Update Data Distribution].
     * Raise a [`MISSING_UPDATE_DATA`] error if the update is not available from either source.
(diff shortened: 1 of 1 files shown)


src/operations/resolve.md
```diff
@@ -165,9 +166,10 @@ Treat [Signal Bytes] as `smt_root`. Look up `smt_root` in `smt_lookup_table` to
     * Raise a [`NOT_FOUND`] error if `resolutionOptions.versionId` is provided.
     * Otherwise, resolve `current_document` as `didDocument`.
 3. Sort `updates` by [BTCR2 Signed Update (data structure)] `targetVersionId` (ascending) with the tuple's block height as a tiebreaker. Remove the first tuple from `updates`.
-4. If `resolutionOptions.versionTime` is provided and the tuple's block `mediantime` {{#cite Bitcoin-Core}} is after `resolutionOptions.versionTime`, resolve `current_document` as `didDocument`. [^4]
-5. Set `block_confirmations` to the tuple's block confirmations.
-6. Set `update` to the tuple's [BTCR2 Signed Update (data structure)] and [check `update.targetVersionId`](#check-update-version).
+4. If `current_document` has no [BTCR2 Beacon] with the tuple's [Beacon Address], ignore the tuple. Continue with the next tuple.
```
danpape, line 169: Would it be worth saying at this point that resolvers MAY remove the tuple from updates to eliminate having to check them over and over again as they continue calling Process Next Update? Leaving them in and ignoring them is fine I suppose, but what happens for a DID with 100s of updates?

parasyte: The algorithm needs to both add and remove updates as it collects new addresses and drops old ones from each updated DID document version. And part of adding new updates is that they need to be sorted.

I don't know of a way to ignore-while-iterating *and* allow arbitrary insertions. This is an example of iterator invalidation. Better to sort the updates into a double-ended queue and consume each from the head. Then the update insertions at arbitrary locations is trivially correct.

danpape: Yeah, I guess we should keep it simple for now. As we construct test vectors for all of our latest spec changes, we could experiment with optimizations and decide if they are worth it.

jintekc: Step 3 removes the tuple from `updates` before step 4 ignores it, so an ignored tuple is not checked again.

danpape: wait, did I totally miss that and start this thread for no reason?! sorry :)

danpape (approved):

peacekeeper (approved):

PR #368, opened 2026-09-14
Specify the funding inputs of a Beacon Signal

### LINKED ISSUE #335: Specify the funding inputs of a beacon announcement transaction
## Problem

The construction of the `OP_RETURN` beacon-signal transaction requires prevouts (UTXOs to spend), a fee, and a change address. The section "Announcing to a Singleton Beacon" ([update.md:126](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/operations/update.md?plain=1#L126)) names none of the three. The spec also does not state whether the library or the caller selects the UTXOs. Therefore the announce APIs of different implementations agree only by chance.

## Proposed Solution

Add a non-normative section. Follow the aggregation precedent at [aggregate-beacons.md:12](https://github.com/dcdpr/did-btcr2/blob/5cd7927/src/beacons/aggregate-beacons.md?plain=1#L12): "A full protocol definition is out of scope for this specification, but a RECOMMENDED example is provided for illustration". In that section, RECOMMEND an announce shape that takes the funding inputs as explicit parameters: prevouts, fee rate, and change address. The caller selects the UTXOs. This design keeps the core sans-I/O: the caller does all network and wallet I/O. The design works for wallet-backed funding and for library-managed funding.

jintekc:
The paragraph ["Announcing to a Singleton Beacon"](https://github.com/dcdpr/did-btcr2/blob/be67bce/src/operations/update.md?plain=1#L129) names three items. They are the hash of the signed update as the Signal Bytes, the signature with the key of the Beacon Address, and the broadcast. A Bitcoin transaction also spends UTXOs, has a fee, and sends the change to an address. The spec does not name these three inputs. It also does not say if the implementation or the caller selects the UTXOs.

This change adds a non-normative subsection under that paragraph. The subsection has the same two parts as the section ["BTCR2 Update Aggregation"](https://github.com/dcdpr/did-btcr2/blob/be67bce/src/beacons/aggregate-beacons.md?plain=1#L12): an out-of-scope sentence, then a RECOMMENDED example. The example is a function with four parameters: the signed update, the prevouts, a fee rate, and a change address. Its output is an [Unsigned Beacon Signal](https://github.com/dcdpr/did-btcr2/blob/be67bce/src/terminology.md?plain=1#L255), a term of the Terminology chapter.

Two decisions in the example are not in the issue. The fee parameter is a rate and not a total. The total changes with the transaction weight, and the caller does not know the weight before the construction. The output is unsigned, because the spec does not give the location of the beacon key. The signature and the broadcast are in the paragraph above the subsection.

Fixes #335.


src/operations/update.md
```diff
@@ -129,6 +129,36 @@ Implementations SHOULD verify `update.proof` before they announce the update. Us
...
+Input arguments:
+
+- `signedUpdate`: The [BTCR2 Signed Update] that the [Beacon Signal] announces.
+- `prevouts`: The [UTXOs][UTXO] that the transaction spends, each with its value. One of them is a [UTXO] that the [Beacon Address] controls.
+- `feeRate`: The fee rate that the caller selects for the transaction.
+- `changeAddress`: The address that receives the total value of `prevouts` minus the fee.
+
+Outputs:
+
+- `unsignedBeaconSignal`: An [Unsigned Beacon Signal]. Its outputs are the change output and then the [Signal Bytes] output.
```
danpape: Sorry, this doesn't really make sense to me. What does "Its" refer to? Are there really two outputs?

jintekc: Ya that's fair. It is poorly worded and and confusing, but "its outputs" is referring to the Unsigned Beacon Signal, which is just an unsigned bitcoin transaction with 2 outputs: change and the OP_RETURN with the signal bytes.

danpape: ... oh, maybe you mean something like "the output of constructBeaconSignal() is an unsigned beacon signal, which is a bitcoin transaction with two outputs: the change output and the [Signal Bytes] output." Or something nicer than that. I guess I was confused by "output" being used in two different ways.

danpape: argh--github didn't show me your reply before I wrote my reply to myself. OK, I get it now.

jintekc: yeah, I agree it needs to change. I've added it to my change log for today. The use of "outputs" in two wildly different contexts that close together is confusing.

jintekc: I simplified it to 

```
- `unsignedBeaconSignal`: An [Unsigned Beacon Signal].
```

because an Unsigned Beacon Signal is defined as "the Bitcoin transaction of the Beacon Signal before its inputs are signed", a Beacon Signal is defined as "a transaction output of the format defined in Signal Bytes" and Signal Bytes is defined as "within the last transaction output". Thus, the link to the definition chain should suffice and restating the definition of a change output is redundant.

danpape (approved):

src/operations/update.md
```diff
@@ -129,6 +129,36 @@ Implementations SHOULD verify `update.proof` before they announce the update. Us
...
+- `signedUpdate`: The [BTCR2 Signed Update] that the [Beacon Signal] announces.
+- `prevouts`: The [UTXOs][UTXO] that the transaction spends, each with its value. One of them is a [UTXO] that the [Beacon Address] controls.
+- `feeRate`: The fee rate that the caller selects for the transaction.
+- `changeAddress`: The address that receives the total value of `prevouts` minus the fee.
+
+Outputs:
+
+- `unsignedBeaconSignal`: An [Unsigned Beacon Signal]
+
+With these parameters, the implementation does not connect to the Bitcoin network and does not use a wallet.
```
parasyte: Grammatical nitpick: Maybe change this to "...does not need a wallet"?

This also appears to be inconsistent with the second paragraph:

> The inputs are [UTXOs][UTXO] that the caller selects, with a wallet or with other software.

"Do we use a wallet, or do we not use a wallet? Or do we use other software?"

parasyte (approved):

src/operations/update.md
jintekc: Yeah, that's fair. It does seem inconsistent and confusing. The two sentences describe two different things.

> The inputs are [UTXOs][UTXO] that the caller selects, with a wallet or with other software.

This describes what the caller of the function does.

> With these parameters, the implementation does not connect to the Bitcoin network and does not use a wallet.

While this describes that the implementation itself does not need external I/O. In this case, that's a Bitcoin node or a wallet. Everything happens offline with the inputs of the function signature.

I think it could be made more clear by saying:

> The inputs are [UTXOs][UTXO] that the caller selects.

And removing the last sentence altogether.

parasyte (approved):

danpape (approved):

